PR #1, opened 2017-06-19
AC-4125: Initial PR for this repository

mc-nathan:
## [AC-4125](https://masschallenge.atlassian.net/browse/AC-4125)

- Defines the subset of former Accelerate models specified in ticket.
- Provides minimal development environment to get the job done.

To test:
```
git clone git@github.com:masschallenge/django-accelerator.git
cd django-accelerator/
git checkout AC-4125
make test
source django-accelerator/venv/bin/activate
make install
cd ..
git clone git@github.com:masschallenge/test-django-accelerator.git
cd test-django-accelerator/mysite
./manage.py migrate
./manage.py shell
from accelerator.models import JobPosting
```


Makefile
```diff
@@ -1,43 +1,49 @@
...
+	@. $(SETUP_ENV); python setup.py sdist
 
 help:
 	@echo "Valid targets are:\n"
 	@for t in $(target_help) ; do \
 	    echo $$t; done
 	@echo
 
-PIP_PACKAGES = Django ipython factory-boy django-mptt
+PIP_PACKAGES = Django django-mptt sorl-thumbnail django-embed-video pillow \
```
shanbady: these should go in a requirements file

shanbady:
I was able to get this running - since I have a feeling I will get some pushback on removing the Makefile altogether, I'm fine with calling this approved and saving that discussion for sometime down the road

mc-nathan:
I agree that the Makefile feels a little bit odd in this case, the key purpose it serves is in providing a place to record how to run migrations and library unit tests.  It's also a nice way to document what is need to package the library.  I'd be open to other suggestions on how to achieve these goals, but it seems like make is a pretty good solution that would be consistent with the other tools we are creating.

Makefile
mc-nathan: Did some research on this and moved most of the requirements into setup.py and then provided a trivial requirements.txt per https://caremad.io/posts/2013/07/setup-vs-requirement/.

It seems like I should be able to do more with .;/setup.py like run tests and maybe migrations, but couldn't get that to turn over, so I left it using the tools derived from Pinax.  I also couldn't get test related requirements to load when running the tests (I think because I'm not using setup.py), so I left a few dependencies in the Makefile related to working on library.

yotammanor:
re @shanbady's comments about using a makefile - in the following link there's a [sample repository](https://github.com/okfn/oki-py) that is used by Open Knowledge International. Their general pattern is that the github repo holds the package they develop _and_ the tools around it that support the dev environment (which don't get packed and uploaded into pypi). This pattern allows for makefile to exist in the repo, and in fact there's a makefile in the sample itself (which I think might be useful for us to look at, [here](https://github.com/okfn/oki-py/blob/master/Makefile)).

yotammanor:
Not a blocker for the moment, but I think the test-django-accelerator should have a different virtual environment that django-accelerator's.

README.rst
```diff
@@ -14,9 +14,64 @@ Quick start
...
-3. The new class should now appear in your application's admin.
+2. Run `python manage.py migrate` in the application that is going to
+manage the models to create the accelerator models.
+
+3. The new class should now appear in your application's admin and through
+the Django shell.
+
+
+Moving Models from Impact API
+-----------------------------
```
yotammanor: Not a blocker, but I don't think this section should be in the main readme.

README.rst
```diff
@@ -14,9 +14,64 @@ Quick start
...
+
+- Remove is_managed import.  Replace call to is_managed with
+  `settings.ACCELERATOR_MODELS_ARE_MANAGED`
+
+- Check to see if there are any weird conditionals and see about
+  adding needed libraries to `PIP_PACKAGES` in the
+  `django-accelerator/Makefile` and any needed applications to
+  `shared_settings.py`
+
+- Run `make clean`
```
yotammanor: make clean is described as "shut down and delete docker images", but this repo does not have any, so I'm not sure what needs to be cleaned here, but something should.

simpleuser/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
```
yotammanor: License Info?

PR #2, opened 2017-06-28
[AC-4501] Keep in sync with impact-api

mc-nathan:


mc-nathan:
This code gets tested in impact-api and is just a copy from there.  Nothing else is dependent on it yet so there isn't a way to more directly test it.

PR #8, opened 2017-08-24
AC-4669: More standard mechanism for tests and making migrations

mc-nathan:
To test:
- Run tests with `make test`
- Change a model and run `make migrations` 
- Install tox with `pip install tox` and run `tox`

Confirm that they all behave as expected.
Review code changes.


Makefile
```diff
@@ -15,11 +15,17 @@ targets = \
 
 target_help = \
   "clean - Shutdown all running containers and removes data files." \
-  "code-check - Runs Flake8 and pep8 on the files changed between the current branch and and a given BRANCH (defaults to development)" \
+  "code-check - Runs Flake8 and pep8 on the files changed between the " \
+  "    current branch and and a given BRANCH (defaults to development)" \
```
pbx: The lines with leading spaces are not indented in the output of `make help`. Not sure what the "right" way is.

mc-nathan: Fixed with \t.

PR #9, opened 2017-08-28
[AC-4788] Adding migration to appease Django 1.8.

mc-nathan:
Django 1.8 seems to want this migration, but Django 1.10 doesn't seem to care one way or another.  This is needed to avoid warnings in accelerate.

### DIFF
--- a/accelerator/migrations/0003_appease_django_18.py
+++ b/accelerator/migrations/0003_appease_django_18.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import re
+import django.core.validators
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accelerator', '0002_avoid_direct_use_of_simpleuser'),
+    ]
+
+    operations = [
+        migrations.AlterField(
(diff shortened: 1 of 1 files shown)


shanbady:
looks like this should be fine since the impact api project doesn't run migrations either way (also it uses unmanaged models so it will ignore schema changes)

mc-nathan:
Full rebuild isn't needed.  To see the problem, simply run `resetdb` on development.  Should report the issue.  To see this change in action, change the rev= for django-accelerator in cfg\base.cfg to AC-4788, run `sudo bin/buildout; resetdb`.  You should no longer get the complaint.

pbx:
Thanks for the update. Unfortunately, `sudo bin/buildout` was not enough to clear out an auto migration I apparently generated while doing review work on August 16th... which happens to be materially the same as the migration here. Doing a full rebuild of my VM and I'll try again.

In any case I still to have the question about why we're adding a field which we recently removed.

mc-nathan:
Good catch regarding the removal of url_slug.  Looks like django-accelerator has not been kept up to date.  Will address.

mc-nathan:
@pbx correctly points out that the startup.py is not current with the Accelerate version.  This should be resolved.  Closing this PR for that.

PR #12, opened 2017-09-06
[AC-4788a] Migration to make Python 2.7 happy

mc-nathan:


accelerator/migrations/0004_update_organization_url_slug.py
```diff
@@ -0,0 +1,35 @@
...
+                blank=True,
+                default='',
+                max_length=64,
+                unique=True,
+                validators=[
+                    django.core.validators.RegexValidator(
+                        '.*\\D.*',
+                        'Slug must contain a non-numeral.'),
+                    django.core.validators.RegexValidator(
+                        re.compile('^[-a-zA-Z0-9_]+\\Z'),
```
pbx: So it look like this migration amounts to removing the `re.UNICODE` flag from this `re.compile` call. How did we discover that was needed?  I'm asking mostly in case it's useful for others to understand as we do work on this Python-version-straddling component.

pbx: (Nathan and I discussed this and for the record the answer is more or less, "that's what Django does when creating this migration under Python 2.7".)

PR #25, opened 2017-10-02
AC-4880 Unify the way we refer to industries in user and organization related objects

pbx:
## [AC-4880](https://masschallenge.atlassian.net/browse/AC-4880)

This ticket changes three repos: accelerate, impact-api, and django-accelerator. 
* [accelerate PR](https://github.com/masschallenge/accelerate/pull/1567)
* [impact-api PR](https://github.com/masschallenge/impact-api/pull/46)

Testing: confirm that everything using the affected fields (now known as `primary_industry` and `additional_industries` functions as before; confirm that UIs use "Primary Industry" and "Additional Industries" when referring to this information.


accelerator/migrations/0005_rename_startup_additional_industry_categories_to_additional_industries.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accelerator', '0004_update_organization_url_slug'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
mc-nathan: I don't think we're relying on this data so it's not nearly as urgent as in the accelerate PR, but this should be RenameField not RemoveField/AddField.

pbx: Fixed.

knittingarch:
@pbx How do I test this in django-accelerate?  There is no UI to check...

knittingarch:
screen shot 2017-10-04 at 9 38 01 am

PR #28, opened 2017-10-05
[AC-4936] Add dependencies for coverage target

mc-nathan:
Retry of PR for AC-4936.


yotammanor:
Setting aside @pbx's comments, `make clean;`make coverage` works for me.
I think it's conceivable for this makefile to assume a python+pip to be installed. But we can add them if we want. Since I think it's optional, I'm approving. @pbx feel free to insist on it.

mc-nathan:
In my view this is a microticket that improves things and is fine to merge.  @pbx if you want pip to get handled or you want to get to the root of your issue with django-admin.py, then I recommend another ticket that is a proper story.

PR #29, opened 2017-10-13
[AC-4975] Update pip to pip3 in Makefile

jonkiparsky:
Testing instructions: 

- review changes, see if they make sense
- try running make targets on your local machine and see if anything breaks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,11 +64,11 @@ DEV_PACKAGES = ipython pep8 flake8 coverage tox \
   factory-boy # factory-boy is in setup.py, but is not getting loaded
 
 $(SETUP_ENV): Makefile requirements.txt
-	@pip install virtualenv
+	@pip3 install virtualenv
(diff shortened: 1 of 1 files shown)


yotammanor:
@jonkiparsky  @mc-nathan Travis for Django-accelerator was setup for both Python2 and Python3. Obviously, now the Python2 env fails. Do we want to remove the 2.7 support for django-accelerator?

mc-nathan:
Hmmm, we should keep the 2.7 check since the accelerate project is using it.  I guess we need to dig into this further since my system is not having this program and it's happily using `pip` rather than `pip3`.

yotammanor:
Well, as far as the MAKEFILE+Travis go, we can parameterize the environment and use pip2/pip3 interchangeably..

jonkiparsky:
Closing this PR, will write a more detailed bug ticket.

PR #31, opened 2017-11-03
AC-4812: Custom User Object Lives Here

mc-nathan:
Work done:
- Implemented the custom user object here
- Added a shared authentication backend that allow the email to come in through either the "email" form field or the "username" form field.
- Switch to Django 1.8 for migration generation
- Bankrupted migrations.

To test: Other than just running the tests.  See the instructions on the corresponding accelerate and impact-api PRs.


yotammanor:
Issues:

- running `make clean;make install` on AC-4812 does not work. On development it does.

settings.py
```diff
@@ -3,6 +3,8 @@
 
 import os
 
+from django import get_version
```
yotammanor: is it left here on purpose?

Makefile
```diff
@@ -57,47 +55,58 @@ help:
...
+VENV = venv
+ACTIVATE = $(VENV)/bin/activate
+
+$(VENV): Makefile requirements.txt
 	@pip install virtualenv
-	@virtualenv -p `which python3` $(ENVIRONMENT_NAME)
-	@touch venv/bin/activate
-	@. venv/bin/activate ; pip install -r requirements.txt; \
+	@rm -rf $(VENV)
+	@virtualenv -p `which python2` $@
```
yotammanor: Unless I change this to `which python2.7` all targets with the `$(VENV)` dependency fail locally.

yotammanor: Also, why are we downgrading to Python2 in the first place? I think we should make reasonable efforts to move towards Python3. django-accelerator itself seems to work well on python3. What caused this change?

simpleuser/models.py
```diff
@@ -31,6 +24,15 @@ def _create_user(self, email, password,
...
             extra_fields["is_active"] = True
+        if "username" not in extra_fields:
+            # For now we need to have a unique id that is at
+            # most 30 characters long.  Using uuid and truncating.
+            # Ideally username goes away entirely at some point
+            # since we're really using email.  If we have to keep
+            # username for some reason then we could switch over
+            # to a string version of the pk which is guaranteed
+            # be unique.
+            extra_fields["username"] = str(uuid.uuid4())[:30]
```
yotammanor: preferably the `30` should be a constant imported from where the limit is being set, but at least a constant defined somewhere. The comment above helps, but still.

Makefile
mc-nathan: All the targets work for me.  There is something different about our environments that we should understand rather than changing this.  The versions of python I have on my system are installed with brew.

We are using Python 2 because the migrations in particular have to be created using Django 1.8 and I don't want to get into mixing Django 1.8 and Python 3 since we don't have to.

settings.py
mc-nathan: Nope.  Removed.  Not sure why make code-check missed that one.

simpleuser/models.py
mc-nathan: Done.

mc-nathan:
Coverage done.

Makefile
yotammanor: (re issues, re-installing python2 through homebrew solved the problem. I'm worried that this is still a future trap for junior developers. I'd add a note in the readme about system prerequisites, and mention there in particular that `which python2` should work. Also possibly a follow on, but one that we should not forget to do)

yotammanor: https://masschallenge.atlassian.net/browse/AC-5049 - `make install3`

yotammanor: https://masschallenge.atlassian.net/browse/AC-5050 - readme with installation instructions

Makefile
```diff
@@ -55,14 +55,14 @@ help:
...
-DEV_PACKAGES = ipython pep8 flake8 coverage tox \
+DEV_PACKAGES = ipython pycodestyle flake8 coverage tox \
   factory-boy # factory-boy is in setup.py, but is not getting loaded
 
 DJANGO_VERSION = 1.8.18
 VENV = venv
 ACTIVATE = $(VENV)/bin/activate
 
-$(VENV): Makefile requirements.txt
+$(VENV): requirements.txt  # Makefile
```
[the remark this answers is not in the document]
mc-nathan: Good catch.  This was a temporary change while I was futzing with some Makefile stuff since having it as a dependency causes the entire environment to rebuild with each change to the Makefile.  I intentionally did it as a comment so if I forgot to uncomment it someone would notice!

PR #36, opened 2018-01-21
[AC-5038-django-accelerator] Switch to Python 3 and Django 1.10.3

mc-nathan:


Makefile
```diff
@@ -58,14 +58,14 @@ help:
 DEV_PACKAGES = ipython pycodestyle flake8 coverage tox \
   factory-boy # factory-boy is in setup.py, but is not getting loaded
 
-DJANGO_VERSION = 1.8.18
+DJANGO_VERSION = 1.10.3
```
yotammanor: I missed the fact that we're on 1.10 in impact-api. We should probably up it to 1.11 as part of the arc in accelerate. Created a ticket.

yotammanor: https://masschallenge.atlassian.net/browse/AC-5279

mc-nathan:
Again, I disagree that there is a blocking issue here.

yotammanor:
@mc-nathan closing this PR, and reopening it on an off-branch, in case we do need to change things on django-accelerator in production during this arc.

PR #39, opened 2018-01-23
[AC-5289] Handle duplicate emails on user authentication

jonkiparsky:
On attempting to authenticate a user, if multiple users w/same email are found, we want to inform the user that it's not them, it's us, and we want to inform developers that we have an opportunity. 

Thus: if MultipleObjectsReturned in EmailModelBackend, we log an `error` with an informative message for developers, and raise an exception. (mcproject handles the exception and shows the helpful message to the user)

To test: see testing steps on related ticket on mcproject.


simpleuser/email_model_backend.py
```diff
@@ -10,10 +14,13 @@ def authenticate(self, email=None, password=None, username=None):
             if not email:
                 email = username
             user = User.objects.get(email=email)
-            if user.check_password(password):
-                return user
         except User.DoesNotExist:
             return None
+        except User.MultipleObjectsReturned:
+            logger.error(MULTIPLE_USERS_FOUND % email)
+            raise User.AuthenticationException
```
yotammanor: I understand why we _might_ want in the future to group a few low-level exceptions under this more generic `AuthenticationException`, but right now it's YAGNI, and adds clutter. Not a lot of clutter, so this is a non-blocker, but I'd consider just raising the built-in `MultipleObjectsReturned` and remove the added Exception.

jonkiparsky: Reusing other people's exceptions is usually a code smell. We're making this shiny and new, we should do it right.

yotammanor: I thought about it more, and I now agree that it makes sense to create a new exception class (given that Django does not specify a relevant Exception for a custom authentication backend). But I also think that it should "belong" to the EmailModelBackend class, not the User class. Does that makes sense to you?

jonkiparsky: I thought about that option, but it seemed to me that User.AuthenticationException made more sense than EmailModelBackend.AuthenticationException. This fits in with the django pattern of attaching the exception to the model. (ie, `SomeModel.MultipleObjectsReturned` instead of `SomeModelManager.objects.MultipleObjectsReturned`) 
Ultimately, it's got to live somewhere and I'm not massively bothered by where it ends up, but this feels more right to me. 
I don't know if this tips the balance for you, but we do not currently have to import EmailModelBackend into accelerate, and that seems like it's a good state to be in. If we move the exception, we have to import the backend.

yotammanor: I have counter arguments, but I don't feel strongly enough to argue them. Approved as it already was.

setup.py
```diff
@@ -57,5 +57,6 @@
     install_requires=INSTALL_REQUIRES,
     tests_require=[
         "factory_boy",
+        "mock",
```
shanbady: havnt tested this but i dont think this needs to be here if we have mock in DEV packages

simpleuser/email_model_backend.py
shanbady: yea this seems odd to me. why are we catching an exception and re-raising a new one? I'd be all for if the  AuthenticationException did something useful but it seems a bit arbitrary. was this introduced to force some kind of behavior?

jonkiparsky: This is because we're trying to write nice shiny code, and recycling other people's exceptions is definitely not that.

setup.py
jonkiparsky: Actually, neither is required. Removing both.

PR #41, opened 2018-02-13
[AC-5039] - Lean Models in Django-accelerator (End-state, no transition)

yotammanor:
**Changes Introduced:**
- All models existing in Union(mcproject, impact-api) were redeclared in django-accelerator
- Declaration of models is in accelerator_abstract, alongside CONSTANTS that are used by the models, and methods that are shared between impact-api and mcproject (lean).
- Concrete subclass is in accelerator, alongside Factories and tests.

**Test:**

- For full functional test of AC-5039, follow instructions on the [impact-api PR](https://github.com/masschallenge/impact-api/pull/107)
- To run tests:
```
git checkout AC-5039
git pull
make test
```


accelerator/tests/test_fb_base_profile.py
```diff
@@ -0,0 +1,16 @@
+# MIT License
```
mc-nathan: Not a blocker, but at some point I think we should get rid of the test_fb_*.py files.  They are now an historic artifact from the days of the first major revision of our factories and the '/fb/' sub-directory is now gone.  In many (most? all?) cases there either isn't a test file without the _fb or the version without is small and easily integrated.  However, this could be handled by a follow on ticket or it could be something we try to address over time.

yotammanor: Removed all test_fb_* from django-accelerator (after saving a few of the tests that were covering code or an interesting functionality)

PR #48, opened 2018-04-02
[AC-5459] Data migration to add mc related permissions

mc-nathan:
To test:
- Get code on development and reload the database:
```
cd accelerate
make checkout
make load-db
make run-server
```
- You should now be able to reproduce the problem with:
-- Masquerade as Ryan: http://localhost:8181/admin/simpleuser/user/47981/masquerade/
-- Goto http://localhost:8181/admin/mc/newsletter/967/
-- At the bottom of the page there should be no cut off "Change" link under "Sections" (see ticket for what it should look like)
-- Goto http://localhost:8181/admin/mc/newsletter/10077/
-- There should be no startup team members at the bottom of the page.

- To see the fix:
```
make stop-server  # Ctrl-C in the server window
cd ../django-accelerator
git checkout AC-5459
cd ../accelerate
make migrate
make run-server
```
- You should no longer see the problem if you go through the above process.


yotammanor:
@mc-nathan 
The instructions are of course incorrect, since I had to change the branch lock for django-accelerator in `requirements/dev.txt` et al (grep for 2018.04.01 ) and then `make build` for the changes to apply. Also the newsletter ID 10077 does not exist on initial schema, but the other change is as described and that's good enough for me.

Merging in. Now we need to release this to take affect.

PR #55, opened 2018-04-09
[AC-5473] Add Texas and Rhode Island to the Office Hours Location menu

knittingarch:
This adds TX and RI to the constants in the model.  After merging, both should show up in the accelerate drop down for available Office Hours locations.


accelerator_abstract/models/base_mentor_program_office_hour.py
```diff
@@ -11,21 +11,25 @@
 from accelerator_abstract.models.accelerator_model import AcceleratorModel
 
 MC_BOS_LOCATION = "MassChallenge Boston"
-MC_CH_LOCATION = "MassChallenge Switzerland"
 MC_IL_JLM_LOCATION = "MassChallenge Israel - Jerusalem"
 MC_IL_TLV_LOCATION = "MassChallenge Israel - Tel Aviv"
 MC_MX_LOCATION = "MassChallenge Mexico"
+MC_RI_LOCATION = "MassChallenge Rhode Island"
```
pbx: I was going to suggest we sort all these, but then I saw that the order follows `LOCATION_CHOICES` below, and we shouldn't just go changing that, so nevermind :)

knittingarch:
@yotammanor Ah, yes, that was the question I asked in general.  I wasn't sure if submitting the PR would somehow trigger a migration.  That was my confusion.  I see now that it works just like the other repos in that regard, just houses the models :)

PR #57, opened 2018-04-23
[AC-5050] - Add prerequisites and installation instructions for the new deployment and build system

yotammanor:


README.md
```diff
@@ -0,0 +1,164 @@
...
+## Quickstart
+
+### As a dependency for MassChallenge's Accelerate/Impact-API
+
+The development environment for Accelerate and [Impact-API](
+https://github.com/masschallenge/impact-api/) were configured to use
+django-accelerator as an editable source.
+
+Follow the instructions on [Setting Up The Development Environment](
+https://github.com/masschallenge/standards/blob/AC-5050/setup_development_environment.md)
```
pbx: It would be nice if we could do this in such a way that it won't 404 once this branch is merged and deleted. Link to master (though it's not there yet)? Don't link? Link just to the top level? Not a blocker.

README.md
```diff
@@ -0,0 +1,164 @@
...
+1.  Add "accelerator" to your INSTALLED\_APPS setting like this:
+    
+        INSTALLED_APPS = [
+            ...
+            'accelerator.apps.AcceleratorConfig',
+        ]
+
+2\. Run `python manage.py migrate` 
+
+3\. The new class should now appear in your application's admin and
```
pbx: New class?

README.md
```diff
@@ -0,0 +1,164 @@
...
+- Python 2.7 + Django 1.8 (to be upgraded soon to 1.11)
+- Python 3.6 + Django 1.10 (to be upgraded soon to 1.11)
+
+Therfore, the code must be compatible for both versions,
+and the Travis builds are configured to test both.
+
+To test this locally, it is possible to run `make tox`.
+
+Also, Running `make test` will simply run the tests
+under the default version.
```
pbx: I think we should either specify what "under the default version" means, or just replace those words with "locally".

README.md
```diff
@@ -0,0 +1,164 @@
...
+be swappable. E.g:
+  ```
+  # accelerator_abstract/models/base_startup.py
+  ...
+    primary_industry = models.ForeignKey(
+        swapper.get_model_name(AcceleratorModel.Meta.app_label, 'Industry'),  
+  ...
+  ```
+- Any use of model that is expected to be swapped, should use
+dynamic imports in a similar fashion. Any code that lives inside
```
pbx: What's meant by "dynamic imports" here? The use of `swapper.get_model_name` as an arg to `ForeignKey` or something else?

README.md
```diff
@@ -0,0 +1,164 @@
...
+   2. [Generating Migrations](#generating-migrations)
+   3. [Running Migrations](#running-migrations)
+   4. [Testing](#testing)
+
+Accelerator is a simple Django package which provides models for the
+MassChallenge Accelerator applications.
+
+## Overview
+
+The package supports two modes of use, in two different django apps:
```
pbx: Cap "Django"

README.md
```diff
@@ -0,0 +1,164 @@
...
+
+3\. The new class should now appear in your application's admin and
+through the Django shell.
+
+## Development
+
+### Changing A Swappable Model
+
+Since Django-accelerator uses [Swappable Models](
+https://github.com/wq/django-swappable-models). changing the django
```
pbx: Cap "Django"

README.md
```diff
@@ -0,0 +1,164 @@
...
+models is slightly different than usual.
+
+**The key differences to note are:**
+
+- All the model declarations - django fields, methods, meta configration,
```
pbx: Cap "Django"

README.md
```diff
@@ -0,0 +1,164 @@
...
+https://github.com/masschallenge/impact-api/) project in
+which this project is installed.
+
+### Testing
+
+This package supports two versions of Django projects:
+- Python 2.7 + Django 1.8 (to be upgraded soon to 1.11)
+- Python 3.6 + Django 1.10 (to be upgraded soon to 1.11)
+
+Therfore, the code must be compatible for both versions,
```
pbx: s/Therfore/Therefore/

README.md
yotammanor: Thanks for noticing that. Changed it to master, we can live with the non-working link until it's merged :)

README.md
yotammanor: changed

README.md
yotammanor: added a clarification

README.md
yotammanor: slightly different. I added an example. I also suggest reading the django swapper docs, they're concise.

yotammanor:
@pbx addressed your comments.

PR #61, opened 2018-05-04
Credit Code redemption report not displaying appropriate results

shanbady:
to test this.

- load up the latest db snaposhot of production.
- checkout development (on django-accelerator)
- visit the refund code redemption report and note that it only shows ~60 refund code redemptions
- checkout this branch (on the django-accelerator project)
- run migrations in impact api ~(run ```make bash-shell``` in impact and then python3 manage.py migrate```~ (run `make migrate` in any of the repos)
- reload the credit code redemption report and note that it loads thousands of redemptions as expected.


accelerator/migrations/0004_fix_acstream_contenttypes.py
```diff
@@ -0,0 +1,181 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.db import (
+    connection,
+    migrations,
+)
+
+
+def fix_actstream_contenttypes(apps, schema_editor):
+    if ContentType.objects.filter(app_label='mc').count() == 0:
+        return
```
yotammanor: Why this condition?

accelerator/migrations/0004_fix_acstream_contenttypes.py
```diff
@@ -0,0 +1,181 @@
...
+    judgeapplicationfeedback_ct = ContentType.objects.get(
+        app_label='mc', model='judgeapplicationfeedback').id
+    startupmentorrelationship_ct = ContentType.objects.get(
+        app_label='mc', model='startupmentorrelationship').id
+    mentorprogramofficehour_ct = ContentType.objects.get(
+        app_label='mc', model='mentorprogramofficehour').id
+    paypalpayment_ct = ContentType.objects.get(
+        app_label='mc', model='paypalpayment').id
+    startup_ct = ContentType.objects.get(
+        app_label='mc', model='startup').id
```
yotammanor: According to `accelerate/mc/apps.py` we are registering a bunch of models on ActStream that do not appear in this report, and probably should. 

Those are the missing lines:

```
registry.register(self.get_model('BaseProfile'))
registry.register(self.get_model('Program'))
registry.register(self.get_model('PayPalRefund'))
registry.register(self.get_model('Panel'))
registry.register(self.get_model('StartupMentorRelationship'))
```

EcoLek:
@yotammanor the number of rows seems low, but the performance being poor is not a surprise. We removed this interface because of how poor the performance was and we haven't addressed that issue.

Also, Test4 is not showing the interface so I don't have a good way of finishing a check of it @shanbady

yotammanor:
@EcoLek try test3?

yotammanor:
@EcoLek both probably got changed when @pbx took over them. Moving to slack.

accelerator/migrations/0004_fix_acstream_contenttypes.py
shanbady: running tests fails without this since the ids are likely different (also. this migration is unnecessary for tests)

accelerator/migrations/0004_fix_acstream_contenttypes.py
shanbady: are you saying that we should not be registering this in apps.py? I believe these are necessary since even though we are not displaying them they do show up as related objects for some of the actions.

yotammanor:
@shanbady had a few suggestions for improvements. In service of Getting It Merged, I created a PR with the suggested fixes: #62

shanbady:
@yotammanor merged your suggested changes. thank you!

PR #69, opened 2018-06-13
[AC-5714] lock Pillow to 5.0.0 because of version issues

jonkiparsky:
Pillow 5.1.0 fails on OSX <10.12 because of XCode problems.
5.1.1 supposedly resolves this but is not available through
pip as yet.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -10,3 +10,4 @@ flake8
 tox
 mock
 factory-boy
+Pillow==5.0.0


mc-nathan:
Unable to test since I cannot reproduce the problem, presumably because I'm on OSX 10.13.

yotammanor:
@mc-nathan what I'd test (or ask @jonkiparsky to demonstrate in this PR) that this change is non-breaking on:
- your machine.
- on travis (=travis builds of impact and accelerate using this branch are green) 
- on opsworks+ECS (=a release using this branch is successful, and does not break things badly. Ideally, use an image field when testing).

Re solving @jonkiparsky's local issues, I guess we'll have to take his word for it :)

mc-nathan:
- My machine works,
- Travis works
- Haven't checked opsworks + ECS, but I have not reason to believe it won't work.

I am a little bit concerned about locking a library on a '.0.0' release in order to support Jon's machine which as far as I know is the only machine we have around that is on 10.11.  I'll bring this up post standup.

mc-nathan:
We're going to hold off on merging this and have asked Jon to upgrade his machine to MacOS 10.13 and see if this is still needed.

PR #70, opened 2018-06-13
[AC-5621] - Create Criterion object for the new Application Allocation system

jonkiparsky:
To test:
check out the branch
run migrations 
run bash-shell on impact-api and examine the db tables for `accelerator_basecriterion` and `accelerator_basecriterionoptionspec`
Should see rows corresponding to the initial options and specs specified in the csv:

```
MySQL [mc_dev]> select * from accelerator_criterion;
+----+----------------------------+----------------------------+----------+----------+------------------+
| id | created_at                 | updated_at                 | type     | name     | judging_round_id |
+----+----------------------------+----------------------------+----------+----------+------------------+
|  1 | 2018-06-14 16:14:54.243759 | 2018-06-14 16:14:54.243813 | reads    | reads    |               48 |
|  2 | 2018-06-14 16:14:54.262810 | 2018-06-14 16:14:54.262965 | matching | industry |               48 |
|  3 | 2018-06-14 16:14:54.265378 | 2018-06-14 16:14:54.265414 | matching | program  |               48 |
|  4 | 2018-06-14 16:14:54.267285 | 2018-06-14 16:14:54.267319 | judge    | role     |               48 |
|  5 | 2018-06-14 16:14:54.270274 | 2018-06-14 16:14:54.270308 | judge    | gender   |               48 |
+----+----------------------------+----------------------------+----------+----------+------------------+
5 rows in set (0.01 sec)

MySQL [mc_dev]> select * from accelerator_criterionoptionspec;
+----+----------------------------+----------------------------+-----------+-------+--------+--------------+
| id | created_at                 | updated_at                 | option    | count | weight | criterion_id |
+----+----------------------------+----------------------------+-----------+-------+--------+--------------+
|  1 | 2018-06-14 16:14:54.258280 | 2018-06-14 16:14:54.258347 |           |     4 |      1 |            1 |
|  2 | 2018-06-14 16:14:54.264250 | 2018-06-14 16:14:54.264292 |           |     1 |      1 |            2 |
|  3 | 2018-06-14 16:14:54.266695 | 2018-06-14 16:14:54.266733 |           |     1 |      1 |            3 |
|  4 | 2018-06-14 16:14:54.268410 | 2018-06-14 16:14:54.268449 | Executive |     2 |      1 |            4 |
|  5 | 2018-06-14 16:14:54.269108 | 2018-06-14 16:14:54.269144 | Investor  |     1 |      1 |            4 |
|  6 | 2018-06-14 16:14:54.269705 | 2018-06-14 16:14:54.269739 | Lawyer    |     1 |      1 |            4 |
|  7 | 2018-06-14 16:14:54.271330 | 2018-06-14 16:14:54.271482 | female    |     1 |      1 |            5 |
|  8 | 2018-06-14 16:14:54.272723 | 2018-06-14 16:14:54.272764 | male      |     1 |      1 |            5 |
+----+----------------------------+----------------------------+-----------+-------+--------+--------------+
8 rows in set (0.01 sec)

MySQL [mc_dev]> select * from accelerator_basecriterionoptionspec;
ERROR 1146 (42S02): Table 'mc_dev.accelerator_basecriterionoptionspec' doesn't exist
MySQL [mc_dev]> select * from accelerator_basecriterion;
ERROR 1146 (42S02): Table 'mc_dev.accelerator_basecriterion' doesn't exist
MySQL [mc_dev]> 
```

Note that this PR does not create any of the helper classes providing functionality for the notional subtypes, that will be done in AC-5622


accelerator_abstract/models/base_criterion.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class BaseCriterion(AcceleratorModel):
+
+    type = CharField(max_length=64)
+    name = CharField(max_length=64)
+    judging_round = ForeignKey(
+        swapper.get_model_name(AcceleratorModel.Meta.app_label,
+                               "JudgingRound"))
+
```
yotammanor: For the swappable mechanism to work well, you need to add to the Base models the following Meta:

```
    class Meta(AcceleratorModel.Meta):
        db_table = '{}_criterion'.format(
            AcceleratorModel.Meta.app_label) . # Note the table name
        abstract = True
```

accelerator_abstract/models/base_criterion_option_spec.py
```diff
@@ -0,0 +1,30 @@
...
+    DEFAULT_COUNT = 1
+    DEFAULT_WEIGHT = 1.0
+
+    option = CharField(max_length=64)
+    count = IntegerField(default=DEFAULT_COUNT)
+    weight = FloatField(default=DEFAULT_WEIGHT)
+    criterion = ForeignKey(swapper.get_model_name(
+        AcceleratorModel.Meta.app_label,
+        "Criterion"))
+
```
yotammanor: likewise missing Meta

accelerator/migrations/0008_auto_20180613_1113.py
```diff
@@ -0,0 +1,90 @@
+# -*- coding: utf-8 -*-
```
yotammanor: the migration module should have a sensible name

accelerator/migrations/0009_auto_20180613_1248.py
```diff
@@ -0,0 +1,64 @@
+# -*- coding: utf-8 -*-
```
yotammanor: likewise sensible module name.

accelerator/migrations/0008_auto_20180613_1113.py
```diff
@@ -0,0 +1,90 @@
...
+        ),
+        migrations.CreateModel(
+            name='Criterion',
+            fields=[
+                ('basecriterion_ptr', models.OneToOneField(auto_created=True, on_delete=django.db.models.deletion.CASCADE, parent_link=True, primary_key=True, serialize=False, to='accelerator.BaseCriterion')),
+            ],
+            options={
+                'abstract': False,
+                'managed': True,
+                'swappable': 'ACCELERATOR_CRITERION_MODEL',
```
yotammanor: 'swappable' should be:
```
                'swappable': swapper.swappable_setting('accelerator',
                                                       'Criterion'),
```

accelerator/migrations/0008_auto_20180613_1113.py
```diff
@@ -0,0 +1,90 @@
...
+        migrations.AlterField(
+            model_name='legalcheck',
+            name='name',
+            field=models.CharField(default='', help_text='Internal name for this check.', max_length=128, unique=True),
+        ),
+        migrations.CreateModel(
+            name='Criterion',
+            fields=[
+                ('basecriterion_ptr', models.OneToOneField(auto_created=True, on_delete=django.db.models.deletion.CASCADE, parent_link=True, primary_key=True, serialize=False, to='accelerator.BaseCriterion')),
+            ],
```
yotammanor: this relationship here appears because BaseCriterion is not abstract.

accelerator/migrations/0008_auto_20180613_1113.py
```diff
@@ -0,0 +1,90 @@
...
+        migrations.AddField(
+            model_name='basecriterionoptionspec',
+            name='criterion',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to=settings.ACCELERATOR_CRITERION_MODEL),
+        ),
+        migrations.AddField(
+            model_name='basecriterion',
+            name='judging_round',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to=settings.ACCELERATOR_JUDGINGROUND_MODEL),
+        ),
```
yotammanor: the value of `to=` should be
```
 to=swapper.get_model_name('accelerator', 'JudgingRound')),
```

accelerator/migrations/0008_auto_20180613_1113.py
jonkiparsky: Hm. Seems like we'd moved away from caring about that. Okay, will fix.

jonkiparsky:
@yotammanor I think I've addressed your concerns. Thanks!

accelerator/tests/factories/__init__.py
```diff
@@ -13,14 +13,16 @@
...
+from .criterion_option_spec_factory import CriterionOptionSpecFactory
 from .currency_factory import CurrencyFactory
 from .entrepreneur_factory import EntrepreneurFactory
 from .entrepreneur_profile_factory import EntrepreneurProfileFactory
 from .expert_category_factory import ExpertCategoryFactory
 from .expert_factory import ExpertFactory
 # Late Loading Factories
 # These fail if put in alphabetical order in the above list
-# presumably due to their SubFactorys.
+# presumably due to their SubFactoriess.
```
mc-nathan: Personally I prefer to always pluralize classes as <classname>s.  It can look a bit funny, but I think it's better if it actually contains the classname and I think it's pretty obvious why it's getting done.  In any case, the double 's' at the end shouldn't be there.  Didn't just fix this since I'd rather leave it to you to decide which fix you think is better.

accelerator_abstract/models/base_criterion.py
```diff
@@ -0,0 +1,29 @@
...
+)
+
+from accelerator_abstract.models.accelerator_model import AcceleratorModel
+
+
+class BaseCriterion(AcceleratorModel):
+
+    type = CharField(max_length=64)
+    name = CharField(max_length=64)
+    judging_round = ForeignKey(
```
mc-nathan: Not necessarily a blocker, but I would expect the JudgingRound association to be for the CriterionOptionSpec since the concept of a "gender criterion" is independent of a JudgingRound, but the count and the weight are JudgingRound specific.

jonkiparsky: Interesting notion. I would say that the criterion as a whole is composed of the Criterion plus its associated Options, and that bundle is judging-round specific, and that structure seems less potentially fragile to me. However, for long-term analysis there might be some value in having the criteria share identity across rounds. 
My inclination is to push back, but if you want to spend some time making the case, I could probably be convinced.

mc-nathan: It's a bit fragile anyway we slice it right now.  In my view the Criterion objects are the lowest level objects that CriterionOptionSpecs then reference.  The association with a JudgingRound should be with a set of CriterionOptionSpecs (ie, fully formed criteria) and ultimately those sets should probably get named, versioned and then shared between JudgingRounds.  As written for judging rounds to have different CriterionOptionSpecs, there would have to be both new Criterion objects and new CriterionOpenSpec objects.  With my suggestion there only need to be new CriterionOptionSpecs.

If that doesn't convince you then I'll just merge it.

jonkiparsky: I say just merge it, as soon as I back out the version lock

PR #84, opened 2018-08-07
[AC-5778] add a reverse reference from mentor to startup mentor relationships

mkiterian:
#### Changes:
- Added a reverse relation name to mentor field

#### Notes:
- These changes are required for `AC-5778` on the impact-api and `AC-5778` on the directory to work


accelerator_abstract/models/base_startup.py
```diff
@@ -183,6 +183,24 @@ def public_inquiry_email(self):
     def public_inquiry_email(self, public_inquiry_email):
         self._set_org_attr("public_inquiry_email", public_inquiry_email)
 
+    @property
+    def mentored_program_location(self):
+        if hasattr(self, "startupmentortrackingrecord_set"):
+            return self.startupmentortrackingrecord_set.first() \
+                       .program.location.split(',')[0]
```
shanbady: for all of these we will probably want to check that self.startupmentortrackingrecord_set.count() is > 0. otherwise the call to .program will fail since first() will return empty

accelerator_abstract/models/base_startup.py
```diff
@@ -183,6 +183,27 @@ def public_inquiry_email(self):
...
+                self.startupmentortrackingrecord_set.count() > 0):
+            return self.startupmentortrackingrecord_set.first() \
+                       .program.location.split(',')[0]
+
+    @property
+    def mentored_program_year(self):
+        if (hasattr(self, "startupmentortrackingrecord_set") and
+                self.startupmentortrackingrecord_set.count() > 0):
+            return self.startupmentortrackingrecord_set.first() \
+                       .program.start_date.year
```
yotammanor: More generally speaking, these methods seem to be mis-located.

I'd expect this property to be on the StartupMentorTrackingRecord, or actually on the StartupMentorProgramRelationship model, and get the location for the exact mentor-program pair (regardless of tracking record, but with regards to the fact that a startup can have more than one mentor, and a mentor can have more than one startup).

accelerator_abstract/models/base_startup.py
```diff
@@ -183,6 +183,27 @@ def public_inquiry_email(self):
     def public_inquiry_email(self, public_inquiry_email):
         self._set_org_attr("public_inquiry_email", public_inquiry_email)
 
+    @property
+    def mentored_program_location(self):
+        if (hasattr(self, "startupmentortrackingrecord_set") and
+                self.startupmentortrackingrecord_set.count() > 0):
+            return self.startupmentortrackingrecord_set.first() \
```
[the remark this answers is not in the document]
mkiterian: On the ticket, to be displayed on the card is:
The location and year are tied to the program to which the mentored startup was in so that they are displayed on the cards for each startup. Refer to https://www.figma.com/proto/QcHVLvEq9PPVnww7u2gtpilq/Profile-Pages?node-id=1%3A2&viewport=738%2C180%2C0.175325&scaling=min-zoom&redirected=1

accelerator_abstract/models/base_startup.py
mkiterian: Graphene works by mapping a model to graphql object type. To be able to access these properties for a startup we have to either add it to the startup model, or create a Mentee subclass and place these methods there

jonkiparsky:
@mkiterian I'm a little curious about the simpleuser migration here, which doesn't relate to a model change in this PR. Is this needed?

jonkiparsky:
Also, please give migration 0016 a descriptive name.

mkiterian:
These migrations are not needed. Closing the PR

mkiterian:
This is not needed because the startup mentor relationships can be accessed through startupmentorrelationship_set

PR #85, opened 2018-08-07
[AC-5894] Explicitly use python3.6 in Makefile

VincentHokie:
#### Changes introduced in [AC-5894](https://masschallenge.atlassian.net/projects/AC/issues/AC-5894):
- Explicitly use python3.6 when running `make migrations`
- Add check and useful error message if python3.6 isn't installed

Background:

It was discovered that as much as the Makefile was pointing to `python3`, it may not be pointing to the correct version of python3 i.e. v3.6. Hence the change and extra check.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,11 @@ VENV = venv
 ACTIVATE_SCRIPT = $(VENV)/bin/activate
 ACTIVATE = export PYTHONPATH=.; . $(ACTIVATE_SCRIPT)
 DJANGO_ADMIN = $(VENV)/bin/django-admin.py
+PYTHON_VERSION = python3.6
+ifeq ($(TRAVIS_PYTHON_VERSION), 2.7)
+	PYTHON_VERSION = python2.7
+endif
+
 
 $(VENV): requirements/base.txt requirements/dev.txt Makefile
 	@if [ ! python3.6 ]; then \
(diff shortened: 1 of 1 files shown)


yotammanor:
@VincentHokie your builds are failing

VincentHokie:
Thanks for the help! The build is all good now @yotammanor

jonkiparsky:
Why python 3.6? Current version is 3.7

VincentHokie:
I can comment on why not v3.5 as well, because it broke when I tried it locally. As far as v3.6, this is what is currently documented and tested against in Travis as well. Explicitly going to 3.7 should probably be first ticketed, researched (and tested) and then implemented (and tested again)

Makefile
```diff
@@ -96,11 +96,19 @@ VENV = venv
...
 ACTIVATE = export PYTHONPATH=.; . $(ACTIVATE_SCRIPT)
 DJANGO_ADMIN = $(VENV)/bin/django-admin.py
+PYTHON_VERSION = python3.6
+ifeq ($(TRAVIS_PYTHON_VERSION), 2.7)
+	PYTHON_VERSION = python2.7
+endif
+
 
 $(VENV): requirements/base.txt requirements/dev.txt Makefile
+	@if [ ! python3.6 ]; then \
```
yotammanor: I'd expect this check to be on `PYTHON_VERSION` directly - so when PYTHON_VERSION is equal `python2.7`, it will test that this version is installed, and not python3.6 (and of course echo an error message for the correct variable as well)

VincentHokie:
made the condition dynamic based on the `PYTHON_VERSION` variable @yotammanor

PR #86, opened 2018-08-08
[AC-5896] - Bug - Make sure `get_profile()` is available on impact-api

yotammanor:
**Changes Introduced in [AC-5896](https://masschallenge.atlassian.net/browse/AC-5896):**
- add get_profile to user class on accelerator app as well.

**Test:**
- in impact-api, run `make django-shell`
run:
```
from accelerator.models import StartupMentorRelationship
StartupMentorRelationship.objects.first()
```
- on django-accelerator's development this fails, and succeeds on branch

### DIFF
--- a/accelerator/apps.py
+++ b/accelerator/apps.py
@@ -6,3 +6,7 @@ from django.apps import AppConfig
 
 class AcceleratorConfig(AppConfig):
     name = 'accelerator'
+
+    def ready(self):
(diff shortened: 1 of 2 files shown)


VincentHokie:
Is there value-add in leaving the `ready` method in the abstract definition? It looks like we're defining the same method in both the abstract and concrete definitions.

yotammanor:
@VincentHokie you are right, I think there's non. Removed it.

PR #87, opened 2018-08-08
[AC-5898] Eliminate naive datetime

jonkiparsky:
Converting the `datetime` to a `date` strips tz info, causing a naive datetime to be created incorrectly.  
Came across this while working on a test for AC-5836, so fixed it. 

To test: 
check out impact-api, run tests. Note timezone warning
point impact-api at this branch, build, and run tests again. Note, no timezone warning

### DIFF
--- a/accelerator/tests/factories/factory_utils.py
+++ b/accelerator/tests/factories/factory_utils.py
@@ -28,5 +28,5 @@ def expert_data(user, password="password"):
         "public_website_consent": True,
         "username": user.username,
         "password": password,
-        "date_joined": user.date_joined,
+        "date_joined": user.date_joined.date(),
     }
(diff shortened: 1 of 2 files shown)


jonkiparsky:
@yotammanor It seems very odd that this change would have caused these issues. Any theories about what's happening here?

jonkiparsky:
Digging in, it looks like the ExpertMutationForm is coming back invalid after this change. 
Looking further.

jonkiparsky:
Verified that this is a testing issue, functionality works on development with this change

jonkiparsky:
@yotammanor Found a way around the problem.

accelerator/tests/factories/factory_utils.py
```diff
@@ -28,5 +28,5 @@ def expert_data(user, password="password"):
         "public_website_consent": True,
         "username": user.username,
         "password": password,
-        "date_joined": user.date_joined,
+        "date_joined": user.date_joined.date(),
```
yotammanor: 😱

PR #90, opened 2018-08-20
[AC-5900] Update contexts

jonkiparsky:
This PR improves contexts to allow some tests on AC-5900 to be written with less fuss and bother.


pbx:
I gave this a try and got blocked by a dev environment bug: https://masschallenge.atlassian.net/browse/AC-5942

PR #91, opened 2018-08-29
[AC-5853] Change context and StartupProgramInterestFactory

jonkiparsky:
Two changes here, a specific change to the context needed for AC-5853 (setting initial interest to 1) and a more general change to StartupProgramInterestFactory so that interest_level will always have an integer value by default. 

I have run tests on accelerate to verify that this change does not break any tests there. 

Note that this PR is a prerequisite for the tests in https://github.com/masschallenge/impact-api/pull/205


pbx:
@jonkiparsky since the branch still exists and this PR is still open, how about cherry-picking [the fix](https://github.com/masschallenge/impact-api/pull/206/commits/d6460bf1e36408aa1cc8144af40bdde7740facce) into this branch? Then I can approve and close and the story will be easier to follow.

PR #94, opened 2018-09-14
[AC-6001] Add allow_dynamic_allocation field to JudgingRound

jonkiparsky:
To test: look at the new field. Notice that I've updated the factory. Look at the help text and request wording changes.


pbx:
@jonkiparsky - I'm having some local environment problems but wanted to post this question while I work them out -- for this PR, what is the easiest way to execute the migration? Assume a base state of fresh clones of all repos and no VMs built.

jonkiparsky:
I just did a `make migrate` - is that not working right for you?

PR #97, opened 2018-09-27
[AC-5989] Programs have a way to make distinctions about if they are included in alumni programs or not

mkiterian:
#### Changes introduced in [AC-5989](https://masschallenge.atlassian.net/browse/AC-5989):
- Add a boolean field to the BaseProgram model to represent alumni eligibility that defaults to False

#### How to test
- Unit test added for the change
- Pull this branch and switch to it
- Run accelerate and go to the url `http://localhost:8181/admin/mc/program/{program_id}/change/`. You should see a checkbox field to with the label `Alumni eligible program`

#### Screenshots
screen shot 2018-09-26 at 21 36 43


accelerator/migrations/0017_add_alumni_program_eligibility.py
```diff
@@ -0,0 +1,35 @@
...
+            model_name='program',
+            name='alumni_eligible_program',
+            field=models.BooleanField(default=False,
+                                      help_text=(
+                                          'Finalists will be added ' +
+                                          'to our Global Alumni ' +
+                                          'Program upon this ' +
+                                          'program being set to "Ended"')),
+        ),
+        migrations.AlterField(
```
jonkiparsky: This migration is from a separate ticket, suggests that there's some confusion of branches here, or possibly that you hadn't run the migrations before starting work on this ticket.

accelerator/tests/test_program.py
```diff
@@ -22,3 +22,7 @@ def test_family_abbr(self):
         program = ProgramFactory(program_family=program_family)
         self.assertEqual(lower_case_slug.upper(),
                          program.family_abbr())
+
+    def test_default_alumni_eligibity_program_property_is_set(self):
```
jonkiparsky: I do not understand what this test is testing or why. It looks like it's testing the ProgramFactory, which we don't need to do.

accelerator_abstract/models/base_program.py
```diff
@@ -57,6 +57,11 @@ class BaseProgram(AcceleratorModel):
         max_length=64,
         choices=PROGRAM_STATUSES,
     )
+    alumni_eligible_program = models.BooleanField(
+        default=False,
+        help_text=("Finalists will be added to our Global " +
+                   "Alumni Program upon this program being set to \"Ended\"")
```
jonkiparsky: Use single quotes around the quoted string if you want to enclose a literal double-quote.

mkiterian:
@jonkiparsky , the unrelated migration has been removed, although it looks like a model was altered without a migration
[AC-5990](https://masschallenge.atlassian.net/browse/AC-5990) Addresses this logic and does not have a impact on how this works, as it's just a toggle

jonkiparsky:
@mkiterian One last concern: it seems to me that we should add a data migration setting this value to True for the programs which we think should be alumni programs. It would also be possible to do this later, but it might as well happen here, while we're creating the field.

mkiterian:
@jonkiparsky , I believe this is a deliverable for [AC-5990](https://masschallenge.atlassian.net/browse/AC-5990) which I'm currently working on, i.e using this value, assign the appropriate Alumni Program Startup Status(startups) and the Alumni Program Role Grant(team members)

jonkiparsky:
Okay, I've added this explicitly to the definition of done for AC-5990

PR #104, opened 2018-10-11
AC-6076: Two leaf migrations present on django-accelerator

VincentHokie:
**Changes introduced in [AC-6076](https://masschallenge.atlassian.net/browse/AC-6076)**:
- add a merge migration to ensure the two lead nodes don't cause a crash during a deploy


pbx:
Just noting here that we agreed in Slack just now to an alternate approach -- renaming `0020_remove_is_open_from_program_family` to `0021_...` and changing its dependency to `0020_criterion_verbose_names`.

That way we reduce mess in the migrations, instead of shipping both a bad-name migration and a merge migration that papers over it.

VincentHokie:
Done the rename and repoint to a different parent.

PR #105, opened 2018-10-12
AC-6079: Refactor accelerate to use `is_employee` as a function from base_utils on django-accelerator

VincentHokie:
**Changes introduced in [AC-6079](https://masschallenge.atlassian.net/browse/AC-6079)**:
- add 'is_employee' function to django-accelerator base_user_utils

### DIFF
--- a/accelerator_abstract/models/base_user_utils.py
+++ b/accelerator_abstract/models/base_user_utils.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2017 MassChallenge, Inc.
 
 from __future__ import unicode_literals
+from accelerator_abstract.models import BaseUserRole
 
 
 def is_expert(user):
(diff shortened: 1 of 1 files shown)


accelerator_abstract/models/base_user_utils.py
```diff
@@ -20,3 +21,10 @@ def _has_user_type(obj, user_type):
     return (obj and
             getattr(obj, "baseprofile", None) and
             obj.baseprofile.user_type == user_type)
+
+
+def is_employee(user):
+    return (not user.is_anonymous and (user.is_superuser or
```
jonkiparsky: In general, I would prefer to avoid complex booleans of this sort. 
I would have broken this up, more like the following:

```
if user.is_anonymous:
  return False
if user.is_superuser:
  return True
return user.programrolegrant_set.filter(program_role__user_role__name=BaseUserRole.STAFF).exists()
```

Not a blocker, info only.

VincentHokie: Thanks for this, integrated suggestion into the changes.

PR #107, opened 2018-10-20
AC-6037: Add PanelTime and PanelSlot collision fields

VincentHokie:
**Changes introduced in [AC-6037](https://masschallenge.atlassian.net/browse/AC-6037)**:
- added `panel_time_collision` and `panel_slot_collision` to judging round model


accelerator_abstract/models/base_judging_round.py
```diff
@@ -206,6 +206,20 @@ class BaseJudgingRound(AcceleratorModel):
...
         help_text='Label for Confirmed Judges',
         related_name='rounds_confirmed_for')
+    panel_time_collision = BooleanField(
+        default=False,
+        verbose_name='Panel Time',
+        help_text=(
+            'Allow us to add a start-up to a PanelTime '
+            'period only once in the active scenarios of this'
+            'judging roound'))
+    panel_slot_collision = BooleanField(
```
jonkiparsky: Not clear to me why we'd want two booleans for this, when the ticket asks for a single boolean choice. If we're anticipating multiple states we probably want an enumeration of options rather than multiple independent fields

accelerator_abstract/models/base_judging_round.py
```diff
@@ -206,6 +206,20 @@ class BaseJudgingRound(AcceleratorModel):
...
+        default=False,
+        verbose_name='Panel Slot',
+        help_text=(
+            'Allow us to add a start-up to a PanelTime '
+            'period more than once in the active scenarios of this'
+            'judging roound'))
```
jonkiparsky: typo

accelerator_abstract/models/base_judging_round.py
```diff
@@ -206,6 +206,20 @@ class BaseJudgingRound(AcceleratorModel):
         null=True,
         help_text='Label for Confirmed Judges',
         related_name='rounds_confirmed_for')
+    panel_time_collision = BooleanField(
+        default=False,
+        verbose_name='Panel Time',
```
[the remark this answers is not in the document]
VincentHokie: Agreed, my first UI consideration was putting in [fieldsets](https://docs.djangoproject.com/en/1.9/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fieldsets) with a general description, then more granular info for the fields themselves. I'll see how it shapes up locally.

accelerator_abstract/models/base_judging_round.py
VincentHokie: added a comment [here](https://masschallenge.atlassian.net/browse/AC-6037) for documentation purposes.

The ticket assumes we need only one BooleanField (for PanelSlot collision) with the assumtion that we have PanelTime collision already, which we don’t.
The two Booleans are therefore for
1. Allowing staff explicitly state that an application can appear in active Scenarios more than once (PanelTime collision) and
2. Allowing staff explicitly state that an application can appear in the same PanelTime more than once (PanelSlot collision)

jonkiparsky:
@VincentHokie I also notice that the judging round factory is not updated. Any time we change a model, we should update the corresponding factory.

VincentHokie:
ah, thanks for the heads-up.

accelerator_abstract/models/base_judging_round.py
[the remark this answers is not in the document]
EcoLek: @VincentHokie This looks a bit confusing to me. What happens when I don't have either one selected? I think we should have specced this one a little more clearly before you picked it up, but I think there's a good direction to go in here.

From my understanding, there are three states possible for collision detection:
* Scenario Level
* Panel Time Level
*. Panel Slot Level

These are how we define things in the database, but maybe we should call them something different to make it more clear to the users.

I'd like to see this be a radio button list, defaulting to the existing system (Option 1)
Radio-Button - https://stackoverflow.com/questions/22514254/django-admin-radio-buttons

Title: Startup scheduling is restricted to:
Option 1: a single pitch for a judging round. (ie. Tuesday AM only)
Option 2: multiple pitches but at different sessions. (ie. Tuesday AM and Tuesday PM)
Option 3: multiple pitches in the same session. (ie. Tuesday AM at Slot 1 and Tuesday AM at Slot 3)

accelerator_abstract/models/base_judging_round.py
```diff
@@ -86,6 +86,18 @@
                       for i in range(9)])
 START_MUST_PRECEDE_END_ERROR = 'Start date must precede end date'
 
+SCENARIO_DETECTION = 'scenario'
+PANEL_TIME_DETECTION = 'panel_time'
+PANEL_SLOT_DETECTION = 'panel_slot'
+COLLISSION_DETECTION_CHOICES = (
```
jonkiparsky: Same typo here, "COLLISSION" should be "COLLISION"

VincentHokie: :)

VincentHokie: fixed.

PR #108, opened 2018-10-23
AC-5946: Add "Time in UTC" in company overview times for programs

VincentHokie:
**Changes introduced in AC-5946**:
- add 'Time in UTC' to overview deadline and start dates on the program model

**How to test**:
- ensure your migration history is linear i.e. the parent of the migration on this PR exists, if it doesn't exist, point it to the last run migration on your local
- run `make migrate`
- On accelerate, visit http://localhost:9191/admin/mc/program/39/change/, see that the overview deadline and start dates have the added `Time in UTC` text to alleviate confusion

- On development, we do not show the user that the time is in UTC.


accelerator_abstract/models/base_program.py
```diff
@@ -120,8 +120,12 @@ class BaseProgram(AcceleratorModel):
         swapper.get_model_name(AcceleratorModel.Meta.app_label, "NamedGroup"),
         blank=True,
         null=True)
-    overview_start_date = models.DateTimeField(blank=True, null=True)
-    overview_deadline_date = models.DateTimeField(blank=True, null=True)
+    overview_start_date = models.DateTimeField(
+        blank=True, null=True,
+        verbose_name="Overview start date (Time in UTC)")
```
pbx: Typically we use `help_text` to add information of this sort, and use `verbose_name` just to improve on the names Django generates from the attribute name. Let's do that here too.

See [this example from the RefundCode model](https://github.com/masschallenge/django-accelerator/blob/development/accelerator_abstract/models/base_refund_code.py#L32) to see both in use.

VincentHokie: changes = made.

VincentHokie: I hope I addressed your concerns as expected

VincentHokie:
Changes = made.

PR #120, opened 2018-12-10
[AC-6187] Change help texts per mockup

jonkiparsky:
Some field names were changed (to protect the innocent). 

Note that we don't check line length on auto-generated migrations, so the travis failures are to be ignored.


accelerator_abstract/models/base_core_profile.py
```diff
@@ -50,17 +50,17 @@ class BaseCoreProfile(AcceleratorModel):
...
     facebook_url = models.URLField(
-        verbose_name="Personal Facebook URL",
+        verbose_name="Facebook",
         blank=True)
     twitter_handle = models.CharField(
-        verbose_name="Personal Twitter Handle",
+        verbose_name="Twitter",
         max_length=16,
         blank=True)
     personal_website_url = models.CharField(
```
pbx: Predates this ticket obviously, but I'm curious why this one field is a `CharField` that uses `url_validator`, rather than a `URLField` like the others.

jonkiparsky: I don't actually know the answer to that, and while I agree that it smells funny, I'm personally not moved to investigate it. 
Or rather, I'm not moved to put aside any of the things I've currently got in the air in order to investigate it. I'm not sure it warrants a ticket until some further digging is done, but if you think that ticket is useful I'll be happy to groom it. 

My suggestion: If you would like to add a code-smell comment here (eg `# we don't know why this is a CharField -pbx 12/17/2019` or something like that) please feel free to do so. That will at least call attention to the oddity and make it more likely that it'll be addressed at some point.

pbx:
Note for QA-hat reviewer: before merging, [look at the migrations](https://github.com/masschallenge/django-accelerator/tree/development/accelerator/migrations) to make sure that no other 0028 migration has been merged. AC-6264 has one in flight.

accelerator_abstract/models/base_core_profile.py
pbx: Done. https://masschallenge.atlassian.net/browse/AC-6299

jonkiparsky:
Looks like AC-6264 is still open, so merging

PR #122, opened 2018-12-11
[AC-5406] changed magic-string keys into pseudo-constants

pbx:
See [accelerate PR](https://github.com/masschallenge/accelerate/pull/1988) for the whole story.

PR #124, opened 2018-12-17
[AC-6264] Criterion Option Spec "option" field cannot be set to "blank" in admin

mkiterian:
#### Changes introduced in [AC-6264](https://masschallenge.atlassian.net/browse/AC-6264):
- add blank=True to the the option field in the criterion option spec model

#### How to test
- checkout to this branch and run migrate
- on run accelerate
- navigate to `/admin/accelerator/criterionoptionspec/`
- You should be able to click on save without seeing field is required message

You shouldn't see this:
screen shot 2018-12-17 at 15 24 14


accelerator/migrations/0028_make_criterion_option_spec_option_field_optional.py
```diff
@@ -0,0 +1,20 @@
...
+# Generated by Django 1.11.14 on 2018-12-17 11:52
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accelerator', '0027_modify_video_url_help_text'),
```
pbx: Warning, this PR is in a race with [AC-5406](https://github.com/masschallenge/django-accelerator/pull/122) -- the loser will need to update its dependency.

jonkiparsky:
@mkiterian You will need to update migrations

PR #129, opened 2019-01-14
AC-5981: Create an criterion_option_spec_context class for use when testing application analysis

VincentHokie:
**Changes introduced in [AC-5981](https://masschallenge.atlassian.net/browse/AC-5981)**
- add criterion_option_spec_context


accelerator/tests/contexts/criterion_option_spec_context.py
```diff
@@ -0,0 +1,48 @@
...
+    ProgramCycleFactory,
+)
+
+
+class CriterionOptionSpecContext:
+    def __init__(self,
+                 type="reads",
+                 name="reads",
+                 judging_round=None,
+                 option="InterestinOption"):
```
jonkiparsky: spelling - "InterestinOption" should be "InterestingOption"

PR #130, opened 2019-01-25
AC-6289: Internal Server Error: /startup/86/payment/12/ 

VincentHokie:
#### Changes introduced in [AC-6289](https://masschallenge.atlassian.net/browse/AC-6289)
- make cycle and startup unique together on the startup cycle interest model

#### How to test
- the first step is to make sure that enforcing this constraint actually prevents the error before we wipe out duplicates
- from your local, run the first two code snippets on [this comment](https://masschallenge.atlassian.net/browse/AC-6289?focusedCommentId=39712&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-39712)
  - this will confirm whether you have duplicates (a key/ keys greater than one signifies duplicates exist)

if you have duplicates (and you should :) ), the third snippet will print all the duplicates
- the printed rows will be of the form `[<number-of-similar-cycle-startup-objects>, <startup-id>-<cycle_id>, <the-rest-tell-us-if-duplicate-objects-have-similar-program-interersts>]` e.g. `[2, '4266-6', True]`
- pick the first duplicate you see
- run the following in the django shell to get the id of the user we will mask as and also put the cycle in a state that allows us reproduce the error
```
from mc.models import *
startup = Startup.objects.filter(id=<startup-id>).first()
user_id = StartupTeamMember.objects.filter(startup=startup).first().user.id
print(user_id) # this is the id of the user you will mask as i.e. http://localhost:8181/admin/simpleuser/user/<user_id>/masquerade/
ProgramCycle.objects.filter(id=<cycle_id>).update(applications_open=True)
```

- leave the above django shell session open we'll need it
- mask as the `user_id` we got in the Django-shell session above
- now visiting `/startup/<startup_id>/payment/<cycle_id>/` should trigger the error
- leave this browser tab open as well

- going back to the open django shell session
- running the following will delete the duplicate for that particular startup-cycle pair
```
duplicates = a["<startup_id>-<cycle_id>"][1:] # e.g. a["4266-6"][1:]
for duplicate in duplicates:
    duplicates.delete()
```

- now try and refresh the browser tab you had open/ open `/startup/<startup_id>/payment/<cycle_id>/`, note the error free page that opens

**to test the scripts**
- run `make migrate` to run the script and enforce the constraint
- you could confirm that there are no duplicates with the following on Django-shell
```
from collections import defaultdict
from mc.models import StartupCycleInterest

a = defaultdict(list) 
for sci in StartupCycleInterest.objects.all():                                                                                             
     key = "{}-{}".format(sci.startup.id, sci.cycle.id)       
     a[key].append(sci)

duplicates = defaultdict(int)
for key, val in a.items():
    duplicates[len(val)] += 1

print(duplicates) # note that his dictionary has no keys greater than 1
```

#### Note
- if you run migrations before getting to the `to test the scripts` section, your db will have no duplicates since we running the migration runs a script to remove all duplicates.


accelerator_abstract/models/base_startup_cycle_interest.py
```diff
@@ -23,6 +23,7 @@ class BaseStartupCycleInterest(AcceleratorModel):
                                        'StartupProgramInterest'))
 
     class Meta(AcceleratorModel.Meta):
+        unique_together = ('cycle', 'startup',)
```
pbx: Trailing comma can be omitted.

jonkiparsky:
Typically we would execute the state cleanup as a migration, to ensure that it executes when this code is deployed. 
If we envision running the cleanup scripts separately from the deploy, we should make a microticket (or task, I suppose - not all that bothered either way) and execute that ticket before the deploy.

VincentHokie:
Good morning @jonkiparsky ! There is actually a ticket for creating the script and executing it in the sprint (https://masschallenge.atlassian.net/browse/AC-4785).

Also, I agree on running the script as part of the migration. I know it's been done before, I'll add it to the migration. That way @pbx 's concern is also addressed. Thanks.

jonkiparsky:
Substantially faster du-deeplication process:

```
In [18]: from django.db.models import Min, Count

In [19]: dupe_preserve_keys = acc.StartupCycleInterest.objects.values("startup_id", "cycle_id").annotate(Min("pk"), Count("pk")).filter(pk__count__gt=1).values_list('pk__min', flat=True)

In [20]: singleton_keys = acc.StartupCycleInterest.objects.values("startup_id", "cycle_id").annotate(Min("pk"), Count("pk")).filter(pk__count=1).values_list('pk__min', flat=True)

In [21]: set(dupe_preserve_keys).intersection(set(singleton_keys))
Out[21]: set()

In [22]: preserve = set(dupe_preserve_keys).union(set(singleton_keys))          

In [23]: acc.StartupCycleInterest.objects.exclude(pk__in=preserve).delete()
```

(takes a few seconds locally, not many minutes)

VincentHokie:
Confirmed the script works and added your suggested script to the migration @jonkiparsky , thank you

accelerator/migrations/0031_make_cycle_and_startup_unique_together_on_startup_cycle_interest.py
```diff
@@ -3,6 +3,23 @@
...
+        'accelerator', 'StartupCycleInterest')
+    dupe_preserve_keys = StartupCycleInterest.objects\
+        .values("startup_id", "cycle_id")\
+        .annotate(Min("pk"), Count("pk"))\
+        .filter(pk__count__gt=1).values_list('pk__min', flat=True)
+    singleton_keys = StartupCycleInterest.objects\
+        .values("startup_id", "cycle_id")\
+        .annotate(Min("pk"), Count("pk"))\
+        .filter(pk__count=1).values_list('pk__min', flat=True)
+    set(dupe_preserve_keys).intersection(set(singleton_keys))
```
jonkiparsky: Not a blocker but this line was only included as a check step and is not actually doing anything useful :)

PR #132, opened 2019-01-31
[AC-4325] Add default values for ApplicationQuestion text box fields

kaymation:
[JIRA](https://masschallenge.atlassian.net/browse/AC-4325)

These fields are not supposed to be blank left blank, so i added defaults

I chose `1` for text_box_lines and `500` for text_limit

I chose these value because they seem like good defaults, that in most cases, the person creating the question will not have to change. To the best of my knowledge, having nonzero values in these fields has no effect on creating non-text questions

To test:
- check out this branch
- go to http://localhost:8181/admin/mc/applicationquestion/add/
- default values should exist in the "Text Box Lines" and "Text Limit" fields


VincentHokie:
Please keep an eye on this [PR](https://github.com/masschallenge/django-accelerator/pull/133), it also has a migration. Whoever's PR is left up will need to update the parent of their migration

PR #133, opened 2019-02-01
[AC-5847] Change to Twitter URL in Signup, User Profile, Startup Profile

kitaka:
Instructions available on the related Accelerate PR:
https://github.com/masschallenge/accelerate/pull/2045

It can be merged independently.


accelerator/migrations/0032_auto_20190131_1051.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.18 on 2019-01-31 15:51
```
pbx: Auto-generated migrations should be renamed, e.g. `0032_twitter_handle_max_length_40.py`

(I've updated [database_management.md](https://github.com/masschallenge/standards/blob/master/database_management.md#creating-and-applying-migrations) to reflect this established but previously undocumented practice.)

PR #141, opened 2019-02-15
AC-6272: Update Judge Progress Stats for New Allocator

VincentHokie:
#### Changes introduced in [AC-6272](https://masschallenge.atlassian.net/browse/AC-6272)
add unique together constraint to judge_application_feedback_factory and allow us change application feedback from judge_feedback_context

#### Context
- These changes are introduced to support a test in the [sibling PR](https://github.com/masschallenge/accelerate/pull/2064) on accelerate

#### Note
- This PR should be merged first as it has dependent tests
- This does not need to be merged in any particular order relative to other django-accelerator PRs since it does not introduce a migration


jonkiparsky:
Note: I'm not a big fan of renaming constants in the import. Approving here because it's in test code, but we should be sure we don't get into the habit for non-test code.

PR #147, opened 2019-03-04
[AC-6461] Add side_navigation to ProgramFamily

jonkiparsky:
Simple PR, just adds a field and migration

Field shows up on django-admin automagically. 

Note that this is a microticket split off from AC-6427 in order to simplify the review and prevent merge-order conflicts with migrations

### **NOTE: Ignore failed check, per standards we do not check code style on auto-generated migrations.**


jonkiparsky:
NOTE: Ignore failed check, per standards we do not check code style on auto-generated migrations.

VincentHokie:
>  NOTE: Ignore failed check, per standards we do not check code style on auto-generated migrations.

This is the first time I'm hearing of this, I and all django-accelerator PRs I have reviewed have edited the migration to be in line with our code-check standards. I'm inclined to enforce that same standard here.

I suggest we cut a ticket to ensure code-check does not apply to migrations and enforce code-check standards on migrations until that ticket is done. Otherwise this comment will have to go on all migration related PRs.

jonkiparsky:
This comment has gone on all migrations that I've ever done :)

In the past, we have never felt that fiddling with the code-check in this way has been the most important use of any engineer's time. If you want to cut that ticket, that's fine, but I doubt it'll ever make it into a sprint. 

In any case, this is a long-standing standard. The rationale is simple: nobody will ever read or modify this file, and eventually it'll get squashed and go away, so fiddling with the lines brings no benefit.

accelerator/migrations/0038_programfamily_side_navigation.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('accelerator', '0037_change_help_texts_for_notes_and_other_mentors_fields'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='programfamily',
+            name='side_navigation',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, to=settings.ACCELERATOR_NAVTREE_MODEL),
```
kaymation: any reason for the interesting import scheme re: models?

jonkiparsky: If you mean in base_program_family.py: 
a) we favor explicit member imports over module imports everywhere else in our codebase, so why not here?
b) there are many, many modules called `models`, so `models.X` does not provide any useful information
c) there is only one place where something like BooleanField comes from, so models.BooleanField again provides no useful information
d) removing gratuitous noise makes the code cleaner and easier to read

pbx:
Re the discussion about linting migrations -- while this _is_ a longstanding practice, it is _not_ covered in our standards docs or in any kaizen as far as I can see. So @jonkiparsky @VincentHokie we at least need a ticket to record it, and possibly one to configure our tooling to respect it.

pbx:
(I had a comment on the migration-linting discussion but I'm saving it for post-standup)

PR #152, opened 2019-03-14
[AC-6367] Enable creating of a Judging Round with end date and time as well as start date and time

shakirandagire:
***Changes introduced in AC-6367:***  
Made the end state time and start date time fields required.

**Association**
The PR is associated with this [ PR on accelerate](https://github.com/masschallenge/accelerate/pull/2091). This PR should be merged before its associated PR on accelerate.

***How to test based on the associated PR.***  
When creating a Judging round as an admin using `/admin/mc/judginground/add/`  
Enter all the details required except the end and start date to which you will be prompted to fill the fields as they are required.

When creating a Judging round as a staff using `/staff/configure_judging_round/` enter all the details required except the end and start date to which you will be prompted to fill the fields as they are required.

Also test other edge cases like inputting end date that is less than the start date and an error will emerge.


accelerator/migrations/0038_auto_20190314_1008.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
jonkiparsky: This migration needs an informative name. 
Also, this seems to be in conflict with existing migrations. I would delete this migration and pull in the most recent changes before recreating.

accelerator_abstract/models/base_judging_round.py
```diff
@@ -108,8 +108,8 @@ class BaseJudgingRound(AcceleratorModel):
         help_text='Include startups from all programs in this Program\'s cycle'
     )
     name = CharField(max_length=60)
-    start_date_time = DateTimeField(blank=True, null=True)
-    end_date_time = DateTimeField(blank=True, null=True)
+    start_date_time = DateTimeField(blank=False, null=True)
+    end_date_time = DateTimeField(blank=False, null=True)
```
jonkiparsky: This leaves `null=True`. Is that what we want? Strictly speaking, if null=True then it is not true that "JudgingRounds without an end date and time are handled gracefully by the log" as requested in the ticket. 

Disallowing nulls will make the migration a little more complicated, but might be worth doing.

accelerator/migrations/0038_auto_20190314_1008.py
shakirandagire: Let me work on the name and also update the migrations

accelerator_abstract/models/base_judging_round.py
shakirandagire: @jonkiparsky I had initially put null=False but since it was non nullable it required a default value and since null=True means that at the database level the field can be null, but in the application level it is a required field.
Thanks for the insight on this.
Let me make the relevant changes

shakirandagire:
@jonkiparsky I have resolved all that was highlighted in the above comments.

.gitignore
```diff
@@ -24,6 +24,7 @@ wheels/
 *.egg-info/
 .installed.cfg
 *.egg
+.vscode/
```
mkiterian: this shouldn't be committed

accelerator_abstract/models/base_judging_round.py
```diff
@@ -108,8 +108,8 @@ class BaseJudgingRound(AcceleratorModel):
         help_text='Include startups from all programs in this Program\'s cycle'
     )
     name = CharField(max_length=60)
-    start_date_time = DateTimeField(blank=True, null=True)
-    end_date_time = DateTimeField(blank=True, null=True)
+    start_date_time = DateTimeField(blank=False, null=False, default=None)
```
mkiterian: if null is False, shouldn't the default value be a datetime object?

shakirandagire: @mkiterian it should but when I put the default value as datetime.now and it prepopulates the start and end date fields(when using the admin panel). So to enter any data I would have to first delete the date then enter the required one.

accelerator/migrations/0042_change_end_and_start_date_time_fields.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accelerator', '0041_make_user_role_field_many_to_many'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
jonkiparsky: These fields probably should not be getting deleted

shakirandagire: I have fixed this issue.

PR #156, opened 2019-03-21
AC-6391: Construct and implement the subnav menu typology using django-sitetree and semantic ui

VincentHokie:
#### Changes introduced in [AC-6391](https://masschallenge.atlassian.net/browse/AC-6391)
- add migration to create the startup dashboard subnav

#### How to test
this setup is deployed on test3
- masquerade as [super+jeff](https://test3.masschallenge.org/admin/simpleuser/user/30263/masquerade/)
- see the [dashboard](https://test3.masschallenge.org/startup/9350/dashboard) for one of the startups

#### Note
Also you will notice that the “Mentors & Goals” link may show up but not have an active link. We currently do checks in our template that we cannot do with sitetree integration.


VincentHokie:
closing PR as we wait for feedback around the startup dashboard from Mexico and a final word around order of operations probably next week.

PR #157, opened 2019-03-21
[Ac 6427] Simple toggling of sitetree sidenav

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-6427

Some technical changes to support the real meat of the work, which is on the related PR (https://github.com/masschallenge/accelerate/pull/2099)

Things that happened here:
- changed the foreignkey field to a boolean toggle, because the described implementation was not in sync with the product's vision
- add the "global sidenav alias" to the NavTree model definition so that it's universally visible (at least, visible anywhere that model is visible, which should be enough)
- update the NavTree context to facilitate testing global nav


kaymation:
Like a shepherd He will tend His flock
Isaiah 40:11

PR #167, opened 2019-04-12
AC-5610: Add issue-number in commit message git-hook to django-accelerator

VincentHokie:
#### Changes introduced in [AC-5610](https://masschallenge.atlassian.net/browse/AC-5610)
- add precommit git hooks

#### How to test
- checkout this branch
- run `make clean`. to delete the venv
- run `make code-check`. this simulates a situation where django-accelerator will rebuild it's virtualenv and in the process set up the git hook
- once this is done, add `import pdb; pdb.set_trace()` in a file and stage the file i.e. `git add <file-you-added-pdb-to>` and attempt to commit the file. Note that the commit is blocked as it would be on impact/ accelerate.
- Now make a random, legal change and commit it. Note that the branch name is appended to the commit message like it is on accelerate/ impact

### DIFF
--- a/git-hooks/pre-commit
+++ b/git-hooks/pre-commit
@@ -24,6 +24,6 @@ exec 1>&2
 DEBUG_FUNC=set_trace
 if git diff $against | grep -v '^-' | grep "$DEBUG_FUNC"
 then
-    echo "Commit failed. 'import $DEBUG_LIB' detected."
+    echo "Commit failed. 'pdb $DEBUG_FUNC' detected."
     exit 1
 fi
--- a/git-hooks/pre-commit
+++ b/git-hooks/pre-commit
@@ -1,0 +1,29 @@
+#!/bin/sh
+#
+# An example hook script to verify what is about to be committed.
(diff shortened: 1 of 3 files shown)


git-hooks/pre-commit
```diff
@@ -0,0 +1,29 @@
...
+# If you want to allow non-ASCII filenames set this variable to true.
+# allownonascii=$(git config --bool hooks.allownonascii)
+
+# Redirect output to stderr.
+exec 1>&2
+
+DEBUG_FUNC=set_trace
+if git diff $against | grep -v '^-' | grep "$DEBUG_FUNC"
+then
+    echo "Commit failed. 'import $DEBUG_LIB' detected."
```
kitaka: @VincentHokie the error could be a little more informative, 
`Commit failed. 'import ' detected.`

I wonder where the value of `$DEBUG_LIB` is supposed to come from.

VincentHokie: You are very right, inherited this bug from accelerate 🤦‍♂️

PR #169, opened 2019-04-15
[AC-6580] Enable side navigation menu items to redirect to fluent pages in the sub navigation

mkiterian:
### Changes introduced in [AC-6580](https://masschallenge.atlassian.net/browse/AC-6580)
- Update fluent redirect url

### NOTE
- This has an accompanying PR on accelerate link to be added

### How to test
- Test instructions on the [accelerate PR](https://github.com/masschallenge/accelerate/pull/2129)


VincentHokie:
taking code quality

accelerator/migrations/0050_update_fluent_redirect_url.py
```diff
@@ -0,0 +1,51 @@
...
+    create_startup_dashboard_subnav()
+
+
+def delete_subnav_trees_and_items(apps, schema_editor):
+    delete_directory_subnav()
+    delete_events_subnav()
+    delete_home_subnav()
+    delete_judging_subnav()
+    delete_resources_subnav()
+    delete_startup_dashboard_subnav()
```
VincentHokie: these reverse migrations aren't actually reversing this migration. If we're going to have reverse migrations, I think it should reverse what is being done in the forward migration

mkiterian: yeah, to be fixed

mkiterian: @VincentHokie, I think the best approach here would be to use the forward migration as the reverse migration. The value that is changed is reflected in a different file hence we can't track versions of it. Practically we won't need to go to the previous values

PR #171, opened 2019-04-18
AC-6406: Clean up existing Twitter handles

VincentHokie:
#### Changes introduced in [AC-6406](https://masschallenge.atlassian.net/browse/AC-6406)
- add migration to clean up the twitter handles on the platform

#### How to test
- checkout this branch and `make migrate`
- from the output, you will see that the twitter handles that have been transformed. They will be of the form
    - starts or ends with a `/`
    - has the handle in the form `*twitter.com/handle`
    - have `n/a` as a handle
    - have their handle as `https://twitter.`
    - have a valid handle with a leading `#`
    - have a hashbang at the beginning i.e. `#!/`
    - have a `lang` get variable on the twitter handle


accelerator/twitter_handle_cleanup/entrepreneur_profile_cleanup.py
```diff
@@ -0,0 +1,76 @@
+import re
+
+
+def clean_entrepreneur_profile_twitter_handles(EntrepreneurProfile):
+    # remove leading twitter.com/ from handles
+    profiles = EntrepreneurProfile.objects.filter(
+        twitter_handle__contains="twitter.com/")
+    for ent in profiles:
```
jonkiparsky: We can let the DB handle this. 
Example code:
`profiles.update(twitter_handle=Func(F('twitter_handle'), Value("twitter.com/"), Value(""), function="replace"))`

Looks like there are only three replacements that we want to do:
```
values = ["https://twitter.com/",
    "https://www.twitter.com/"
    "twitter.com/"]
```
So embedding the above example in a loop over that list 

```
for value in values:
    profiles.update(twitter_handle=Func(F('twitter_handle'), Value(value), Value(""), function="replace"))
```

will eliminate a lot of code and save all of us a lot of load-db time.

accelerator/twitter_handle_cleanup/entrepreneur_profile_cleanup.py
```diff
@@ -0,0 +1,76 @@
...
+        old = ent.twitter_handle
+        new = ent.twitter_handle[ent.twitter_handle.index('twitter.com/')+12:]
+        ent.twitter_handle = new
+        ent.save()
+        print(ent, old, new)
```
pbx: Leftover debugging/development `print` statement (sometimes known as a "**pront**" ) -- please remove!

(There are 20, by my count)

accelerator/migrations/0050_cleanup_twitter_urls.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def clean_up_twitter_handles(apps, schema_editor):
+    Organization = apps.get_model('accelerator', 'Organization')
+    ExpertProfile = apps.get_model('accelerator', 'ExpertProfile')
+    EntrepreneurProfile = apps.get_model(
+        'accelerator',
+        'EntrepreneurProfile')
+
+    clean_entrepreneur_profile_twitter_handles(EntrepreneurProfile)
```
pbx: While I like the compact and readable top level of this migration, the supporting files (clean_*_twitter_handles) could use some improvement.

We don't hold migrations to the same standard as application code -- but for something this large (several hundred lines) a bit more DRY/readability effort is in order I think. Certainly as a reviewer I'm unlikely to catch a logic error in the third repetition of a given block of string-munging code.

accelerator/twitter_handle_cleanup/entrepreneur_profile_cleanup.py
pbx: Asante kindly pointed out that the ticket requests this output, so never mind!

accelerator/twitter_handle_cleanup/entrepreneur_profile_cleanup.py
VincentHokie: these are the variations I found

```
[
        '@https://twitter.com/',
        'Http://twitter.com/',
        'Twitter.com/',
        'http://twitter.com/',
        'http://www.twitter.com/',
        'https://mobile.twitter.com/',
        'https://twitter.com/',
        'https://ww.twitter.com/',
        'https://www.Twitter.com/',
        'https://www.twitter.com/',
        'https;//twitter.com/',
        'htttp://www.twitter.com/',
        'maapit or https://twitter.com/',
        'twitter.com/',
        'www.Twitter.com/',
        'www.twitter.com/',
    ]
```

but I get the idea 😄

accelerator/twitter_handle_cleanup/utils.py
```diff
@@ -0,0 +1,153 @@
...
+    F,
+    Func,
+    Value
+)
+
+
+TWITTER_HANDLE_CSV_FILE_NAME = 'twitter_handle_cleanup.csv'
+
+
+def remove_leading_slashes_from_strings(obj):
```
pbx: I am utterly bikeshedding here, but I'd vote for just calling this `remove_leading_slashes`.

accelerator/twitter_handle_cleanup/utils.py
```diff
@@ -0,0 +1,153 @@
+import re
```
pbx: I think it would be a kindness to add a comment with the ticket number, at least in this file. We may never use these functions again, but if somebody's curious about them the ticket number will get them a quick answer.

accelerator/twitter_handle_cleanup/utils.py
```diff
@@ -0,0 +1,153 @@
...
+            obj.id,
+            twitter_handle,
+            new_twitter_handle
+        ]
+        write_to_csv(profile_list)
+        obj.twitter_handle = new_twitter_handle
+        obj.save()
+
+
+def remove_trailing_slashes_from_strings(obj):
```
pbx: As above -- how about `remove_trailing_slashes`.

accelerator/twitter_handle_cleanup/utils.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def remove_trailing_and_leading_whitespace(obj):
+    new_twitter_handle = str(obj.twitter_handle).strip()
+    if new_twitter_handle != obj.twitter_handle:
+        obj.twitter_handle = new_twitter_handle
+        obj.save()
+
+
+def remove_leading_hastag_on_valid_twitter_handles(Model):
```
pbx: Typo, s/hastag/hashtag/

accelerator/twitter_handle_cleanup/utils.py
```diff
@@ -0,0 +1,153 @@
...
+                    F('twitter_handle'),
+                    Value("#"),
+                    Value(""),
+                    function="replace"))
+
+
+def remove_twitter_url_prefix_from_handles(Model):
+
+    twitter_url_variations = [
+        'maapit or https://twitter.com/',
```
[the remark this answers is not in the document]
VincentHokie: yeah, this is a particular case. I was looking for every combination of `*twitter.com/*` to replace and got this as well.

VincentHokie:
Worked on the suggested changes.

jonkiparsky:
@pbx I have added an approve. I'm leaving the merge to you, in case there's any reason to hold off.

PR #174, opened 2019-04-19
AC-6622: AC-6622: Internal Server Error on /startup/000/team/add and /accounts/register/entrepreneur/

VincentHokie:
#### Changes introduced in [AC-6622](https://masschallenge.atlassian.net/browse/AC-6622)
- override create method on objects manager to delete existing profile before creating another

#### How to test
On local
- on accelerate open the `mcproject/settings.py` file and fill in the fields below
```
ALGOLIA = {
    'APPLICATION_ID': '', # get this from your impact-api .env file i.e. ALGOLIA_APPLICATION_ID
    'API_KEY': '', # get this from your impact-api .env file i.e. ALGOLIA_API_KEY
    'INDEX_PREFIX': 'dev'
}

ALGOLIA_ENABLED = True
```
- while on development on all branches
- go to http://localhost:8181/accounts/register/entrepreneur/ and create an entrepreneur, see the failure with a duplicate user error
- checkout AC-6622 on accelerate and accelerator
- go to http://localhost:8181/accounts/register/entrepreneur/ and create an entrepreneur, see that it works

This is also deployed on test2 for further QA testing.

#### Note
Ensure to merge this before the [sibling PR](https://github.com/masschallenge/accelerate/pull/2141)

PR #179, opened 2019-05-08
AC-6662: 500 error received when attempting to delete a user

VincentHokie:
#### Changes introduced in [AC-6662](https://masschallenge.atlassian.net/browse/AC-6662)
- stop logic from auto creating profiles on query set `get`, instead do that on user creation

#### How to test
- with a clean db
- while on development on both accelerate and django-accelerator
- log in as demoadmin
- try and delete [30-user](http://localhost:8181/admin/simpleuser/user/30/delete/) and [38298-user](http://localhost:8181/admin/simpleuser/user/38298/delete/), notice that both of these fail.
    - 38298-user has PartnerTeamMembers associated
    - 30-user has both PartnerTeamMembers and StartupStatus associated
- now checkout AC-6662 on both accelerate and django-accelerator
- try and delete [30-user](http://localhost:8181/admin/simpleuser/user/30/delete/) and [38298-user](http://localhost:8181/admin/simpleuser/user/38298/delete/) again, notice that it now passes. Success🏆 

#### Note
This PR has a sibling on [accelerate](https://github.com/masschallenge/accelerate/pull/2154) and this should be merged in first

### DIFF
--- a/accelerator/models/profile_query_set.py
+++ b/accelerator/models/profile_query_set.py
@@ -74,7 +74,8 @@ class ProfileQuerySet(QuerySet):
     def _profile_for_inferred_profile_type(self, profile_manager):
         if profile_manager is None:
             return None
-        profile = profile_manager.using(self._db).filter(user=self.user).first()
+        profile_qs = profile_manager.using(self._db).filter(user=self.user)
+        profile = profile_qs.first()
         if not profile:
             profile = self._get_profile_from_existing_profile_types()
         return profile
--- a/accelerator/models/profile_query_set.py
+++ b/accelerator/models/profile_query_set.py
@@ -28,9 +28,6 @@ BaseProfile = swapper.load_model(AcceleratorConfig.name,
                                  "BaseProfile")
 User = get_user_model()
 
-MISSING_BASE_PROFILE_TEMPLATE = ("Missing BaseProfile for user {}. "
-                                 "Creating.")
-
 logger = logging.getLogger(__file__)
 
 EMAIL_KEY = 'email'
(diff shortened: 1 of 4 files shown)


pbx:
Taking QA.

accelerator/models/profile_query_set.py
```diff
@@ -56,22 +53,13 @@ def get(self, *args, **kwargs):
...
-            try:
-                self.user = User.objects.get(email=kwargs[EMAIL_KEY])
-            except User.MultipleObjectsReturned:
-                # This guard is necessary since emails are not necessarily
-                # unique in our system.
-                # Should be able to remove this when we define a custom User
-                self.user = User.objects.filter(
-                    email=kwargs[EMAIL_KEY]).first()
+            self.user = User.objects.get(email=kwargs[EMAIL_KEY])
+
```
pbx: Hurrah for deleting old crufty code

settings.py
```diff
@@ -2,6 +2,12 @@
 # Copyright (c) 2017 MassChallenge, Inc.
 
 import os
+import sys
+import logging
+
+if len(sys.argv) > 1 and sys.argv[1] == 'test':
+    logging.disable(logging.CRITICAL)
```
pbx: Why did you add this? Generally `if test` is something we want to avoid.

VincentHokie:
Will continue to address concerns after @jonkiparsky 's concerns are discussed today.

settings.py
VincentHokie: logging was [suggested](https://github.com/masschallenge/django-accelerator/pull/174#discussion_r277315271) when a persons Member profile is deleted in favour of an Expert/ Entrepreneur profile.

Without this logic, our tests are flooded with those log messages.

VincentHokie:
Closing this PR pending a discussion by team around this use case

PR #181, opened 2019-05-13
[AC-5942] Allow django-accelerator Makefile to work with Python 3.7, not just Python 3.6

pbx:
## [AC-5942](https://masschallenge.atlassian.net/browse/AC-5942)

**To test**: confirm that Makefile doesn't abort when called with Python 3.7 (while confirming it still works with Python 3.6).

(If you want help on how to do this locally, let me know.)

Deep Python nerd background on `python` / `python2` / `python3`: [PEP 394](https://www.python.org/dev/peps/pep-0394/)


mkiterian:
Taking code review, will do some QA as well

VincentHokie:
if @mkiterian  is on the right track, then this should probably change as well
https://github.com/masschallenge/django-accelerator/blob/a4123586f3fa90278b222efffc6a944af97540e9/setup.py#L44-L61

VincentHokie:
taking QA here

VincentHokie:
@pbx, you missed a 3.6 => 3.7 change in the .travis.yml file

VincentHokie:
with python3.7 `code-check` produces this warning that qns will arise about.
```
~/mc/django-accelerator [AC-6678] $ make code-check
/Users/vincentasantehokie/mc/django-accelerator/venv/lib/python3.7/site-packages/pycodestyle.py:113: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
```

I think we should resolve this/ add a note to ignore the warning.
I was able to resolve the warning by removing the constraints on the pycodestyle and flake8 [dependencies](https://github.com/masschallenge/django-accelerator/blob/a4123586f3fa90278b222efffc6a944af97540e9/requirements/dev.txt#L8-L9)


Actually, scratch my earlier [comment](https://github.com/masschallenge/django-accelerator/pull/181#issuecomment-493449909), I think travis should run against all major python3 versions since were not specifying which python version to use.

VincentHokie:
My python3 points to 3.5.4 and I get this after a `make clean; make code-check`

```
ERROR: Complete output from command python setup.py egg_info:
    ERROR: Download error on https://pypi.org/simple/nose/: [SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:719) -- Some packages may not be found!
    Couldn't find index page for 'nose' (maybe misspelled?)
    Download error on https://pypi.org/simple/: [SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:719) -- Some packages may not be found!
    No local packages or working download links found for nose
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/f0/5xdtxgk132gg6kztl7zvykb80000gn/T/pip-install-a0jknich/django-embed-video/setup.py", line 37, in <module>
        test_suite='nose.collector',
      File "/Users/vincentasantehokie/mc/django-accelerator/venv/lib/python3.5/site-packages/setuptools/__init__.py", line 144, in setup
        _install_setup_requires(attrs)
      File "/Users/vincentasantehokie/mc/django-accelerator/venv/lib/python3.5/site-packages/setuptools/__init__.py", line 139, in _install_setup_requires
        dist.fetch_build_eggs(dist.setup_requires)
      File "/Users/vincentasantehokie/mc/django-accelerator/venv/lib/python3.5/site-packages/setuptools/dist.py", line 717, in fetch_build_eggs
        replace_conflicting=True,
      File "/Users/vincentasantehokie/mc/django-accelerator/venv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 782, in resolve
        replace_conflicting=replace_conflicting
      File "/Users/vincentasantehokie/mc/django-accelerator/venv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 1065, in best_match
        return self.obtain(req, installer)
      File "/Users/vincentasantehokie/mc/django-accelerator/venv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 1077, in obtain
        return installer(requirement)
      File "/Users/vincentasantehokie/mc/django-accelerator/venv/lib/python3.5/site-packages/setuptools/dist.py", line 784, in fetch_build_egg
        return cmd.easy_install(req)
      File "/Users/vincentasantehokie/mc/django-accelerator/venv/lib/python3.5/site-packages/setuptools/command/easy_install.py", line 673, in easy_install
        raise DistutilsError(msg)
    distutils.errors.DistutilsError: Could not find suitable distribution for Requirement.parse('nose')
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/f0/5xdtxgk132gg6kztl7zvykb80000gn/T/pip-install-a0jknich/django-embed-video/
```

Restricting this to a few python versions may be the way here, allowing travis to run on all python3 versions would give us an idea of where to start with this.

pbx:
Thanks for the thorough testing. My thoughts on the issues raised:
* I agree the `EXTRANEOUS_WHITESPACE_REGEX` warning is distracting, it would be nice to get rid of it. Based on [this bug](https://github.com/PyCQA/pycodestyle/issues/728#issuecomment-394948529) I think it's just pycodestyle (fixed in the version our pin says not to exceed :)
* Python 3.5 is not intended to be supported -- just 3.6 and 3.7. Perhaps we should state that explicitly in the project READMEs.
* Similarly, this ticket is _just_ about enabling engineers to use the current version of Python 3; it is _not_ about changes to Travis.

Happy to discuss these points further as needed.

VincentHokie:
> I agree the EXTRANEOUS_WHITESPACE_REGEX warning is distracting, it would be nice to get rid of it. Based on this bug I think it's just pycodestyle (fixed in the version our pin says not to exceed :)

Yeah, I'll test python 3.6 and python3.7 compatibility without the restriction to our pycodestyle version.

> Python 3.5 is not intended to be supported -- just 3.6 and 3.7. Perhaps we should state that explicitly in the project READMEs.

As well as enforce it in the Makefile so that engineers see an error rather than a failed build.

> Similarly, this ticket is just about enabling engineers to use the current version of Python 3; it is not about changes to Travis.

I still think, though, that running django-accelerators travis is important since engineers will be using it locally..we could potentially catch issues early with an extra 3.7 build

pbx:
@VincentHokie -- if you still have blocking concerns, please post a blocking review and I'll address them. (Also, i'm curious if your pycodestyle version test confirmed my guess!)

VincentHokie:
Doing testing now, @pbx . sorry for the delay

setup.py
```diff
@@ -57,6 +57,7 @@
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
```
VincentHokie: we don't actually support python 3.4 and 3.5. We probably shouldn't have those here

pbx:
@VincentHokie - thanks for catching the `classifiers` problem, fixed. 

Re testing under Python 3.7, we _do_ have a tox target for it...  Perhaps we should set up a periodic run of that, analogous to our daily Travis builds -- to prevent surprises as you note.

Re adding logic to the makefile (if I understand correctly what you are asking for), I'd really rather not. I think we can trust engineers not to arbitrarily change their local Python versions to ones we expressly do not recommend (e.g. 2.6, 3.4), and the Makefile is complex enough already.  If i am misunderstanding, please clarify.

PR #185, opened 2019-05-22
AC-6678: Allow a merge-to-deploy workflow to an AWS Stack e.g pre-staging

VincentHokie:
#### Changes introduced in [AC-6678](https://masschallenge.atlassian.net/browse/AC-6678)
- add logic that triggers an impact build when development is built

#### How to test
- see that the [last build of this branch](https://travis-ci.org/masschallenge/impact-api/builds/541703559), led to an [impact build](https://travis-ci.org/masschallenge/impact-api/builds/535794903) being triggered with the build message being **Triggered by '[AC-6678] remove dummy migration' in the Front-End**
- rerunning that build should lead to the same result, i.e. an impact build being triggered that leads to an ECS deploy

#### Note
- This currently uses a SANTE_GH_TOKEN variable that we `need` to replace. 
   - ~Should I put this in 1Pass and we keep using it?~
   - ~can we generate an MC token not tied to an engineers account?~
   - ~if we can't, whose account do we then use to generate a GH token?~
   - is it time to create a [GH Machine User](https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users) whose ssh keys we'll use for automation --> created, ticket to address this has been created ([AC-6760](https://masschallenge.atlassian.net/browse/AC-6760))

#### TLDR
There's two pieces of follow-on work to this
- [AC-6760](https://masschallenge.atlassian.net/browse/AC-6760)
- [AC-6754](https://masschallenge.atlassian.net/browse/AC-6754)

[AC-6678]: https://masschallenge.atlassian.net/browse/AC-6678


.travis.yml
```diff
@@ -27,3 +27,29 @@ install:
 script:
 - make test
 - make code-check
+
+after_success:
+- >
+  if [ ! "$TRAVIS_PULL_REQUEST" ] && [ "$TRAVIS_BRANCH" = "development" ] || [ "$TRAVIS_BRANCH" = "AC-6678" ]; then
```
kitaka: Similar to the comment on the front-end PR. Is `jq` really necessary?

VincentHokie:
yes, @pbx ...that build was for a commit that doesn't exist anymore. This should work https://travis-ci.org/masschallenge/impact-api/builds/541703559, I'll update the link in the instructions as well

.travis.yml
VincentHokie: sorted now

pbx:
Thanks for filling in the blanks.  Confirmed to work, I approve.

PR #186, opened 2019-05-23
AC-6662: 500 error received when attempting to delete a user

VincentHokie:
#### Changes introduced in [AC-6662](https://masschallenge.atlassian.net/browse/AC-6662)
- stop logic from auto creating profiles on query set `get`, instead do that on user creation
- add migration that revokes user deletion permissions from non superuser users

#### How to test
- with a clean db
- while on development on both accelerate and django-accelerator
- get into the **accelerate** django-shell, i.e. `make django-shell`
- perform the following magic that will show you the count of superusers and non-superusers with user deletion powers
```
from django.contrib.auth.models import Permission
from mc.models import User
perm = Permission.objects.filter(name="Can delete user").first()
User.objects.filter(is_superuser=True).count() # take note of this
User.objects.filter(user_permissions=perm).exclude(is_superuser=True).count() # take note of this
```

- [mask](http://localhost:8181/admin/simpleuser/user/4697/masquerade/) as a user that is not a superuser and has user deletion permission
- try and delete [30-user](http://localhost:8181/admin/simpleuser/user/30/delete/), [50193-user](http://localhost:8181/admin/simpleuser/user/50193/delete/) and [38298-user](http://localhost:8181/admin/simpleuser/user/38298/delete/), notice that both of these fail.
   - 50193-user has PartnerTeamMembers  associated
    - 38298-user has Startup Team Members and StartupStatus associated
    - 30-user has both PartnerTeamMembers and Startup Team Members associated
- now checkout AC-6662 on both accelerate and django-accelerator
- run migrations i.e. `make migrate`
- access the **accelerate** django-shell and type in the following
```
from django.contrib.auth.models import Permission
from mc.models import User
perm = Permission.objects.filter(name=“Can delete user").first()
User.objects.filter(is_superuser=True).count() # should be the same as before
User.objects.filter(user_permissions=perm).exclude(is_superuser=True).count() # should be 0
```

- try and delete [30-user](http://localhost:8181/admin/simpleuser/user/30/delete/) and notice you don't have the **delete** button of the page since your delete permissions have been revoked.
- release this user and be demoadmin
- try and delete [30-user](http://localhost:8181/admin/simpleuser/user/30/delete/), [50193-user](http://localhost:8181/admin/simpleuser/user/50193/delete/) and [38298-user](http://localhost:8181/admin/simpleuser/user/38298/delete/) again, notice that it now passes. Success🏆

#### Note
This PR has a sibling on [accelerate](https://github.com/masschallenge/accelerate/pull/2182) and this should be merged in first

Was convinced by @jonkiparsky that the extent of the cleanup initially was necessary but premature. This should be an easier review :)

### DIFF
--- a/accelerator/migrations/0054_revoke_delete_user_perms_from_non_superuser.py
+++ b/accelerator/migrations/0054_revoke_delete_user_perms_from_non_superuser.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.contrib.auth.models import Permission
+
+
+def revoke_delete_user_perms_from_non_superusers(apps, schema_editor):
+    perm = Permission.objects.filter(name="Can delete user").first()
+    if perm and perm.user_set.exists():
+        perm.user_set.clear()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        (
+            'accelerator',
+            '0053_add_eventbrite_organization_id_field'
+        ),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            revoke_delete_user_perms_from_non_superusers,
+            migrations.RunPython.noop),
+    ]
--- a/accelerator/migrations/0054_revoke_delete_user_perms_from_non_superuser.py
+++ b/accelerator/migrations/0054_revoke_delete_user_perms_from_non_superuser.py
@@ -5,7 +5,7 @@ from django.db import migrations
 from django.contrib.auth.models import Permission
 
 
-def revoke_delete_user_perms_from_non_superusers(apps, schema_editor):
+def revoke_delete_user_perms_from_all_users(apps, schema_editor):
     perm = Permission.objects.filter(name="Can delete user").first()
     if perm and perm.user_set.exists():
         perm.user_set.clear()
@@ -22,6 +22,6 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.RunPython(
-            revoke_delete_user_perms_from_non_superusers,
+            revoke_delete_user_perms_from_all_users,
             migrations.RunPython.noop),
     ]
--- a/accelerator/models/profile_query_set.py
+++ b/accelerator/models/profile_query_set.py
@@ -56,14 +56,8 @@ class ProfileQuerySet(QuerySet):
         if PK_KEY in kwargs.keys():
             self.user = User.objects.get(pk=kwargs[PK_KEY])
         elif EMAIL_KEY in kwargs.keys():
-            try:
-                self.user = User.objects.get(email=kwargs[EMAIL_KEY])
-            except User.MultipleObjectsReturned:
-                # This guard is necessary since emails are not necessarily
-                # unique in our system.
-                # Should be able to remove this when we define a custom User
-                self.user = User.objects.filter(
-                    email=kwargs[EMAIL_KEY]).first()
+            self.user = User.objects.get(email=kwargs[EMAIL_KEY])
+
         profile_manager = self._profile_manager_by_user_type()
 
         profile = self._profile_for_inferred_profile_type(profile_manager)
--- a/accelerator/migrations/0055_revoke_delete_user_perms_from_non_superuser.py
+++ b/accelerator/migrations/0055_revoke_delete_user_perms_from_non_superuser.py
@@ -16,7 +16,7 @@ class Migration(migrations.Migration):
     dependencies = [
         (
             'accelerator',
-            '0053_add_eventbrite_organization_id_field'
+            '0054_add_eventbrite_organizer_id_field'
         ),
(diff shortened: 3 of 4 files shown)


accelerator/migrations/0054_revoke_delete_user_perms_from_non_superuser.py
```diff
@@ -0,0 +1,31 @@
...
+
+from django.db import migrations
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
+
+
+def revoke_delete_user_perms_from_non_superusers(apps, schema_editor):
+    User = get_user_model()
+    perm = Permission.objects.filter(name="Can delete user").first()
+    users = User.objects.filter(user_permissions=perm).exclude(
```
jonkiparsky: Probably quickest and easiest to just do `perm.user_set.clear()`

accelerator/tests/test_get_profile.py
```diff
@@ -94,18 +91,3 @@ def test_get_profile_handles_incorrect_user_type_expert(self, mock_logger):
         mock_logger.warning.assert_called_with(
             INCORRECT_USER_TYPE_TEMPLATE.format(ENTREPRENEUR_USER_TYPE,
                                                 EXPERT_USER_TYPE))
-
-    @patch("accelerator.models.profile_query_set.logger")
-    def test_get_profile_handles_a_user_without_a_profile(self, mock_logger):
```
jonkiparsky: why is this test being removed?

VincentHokie: we no longer auto create a user when the `get_profile` method is called. we do it [on user creation](https://github.com/masschallenge/django-accelerator/pull/186/files#diff-fc9eab3fc52c895c2bce5da5cd20f3e9R52) and therefore have no log warnings for when a user is created like this.

VincentHokie: creating profiles like this counter intuitive and has caused pain e.g. user registration bug, now this.

accelerator/migrations/0054_revoke_delete_user_perms_from_non_superuser.py
VincentHokie: neat!

VincentHokie: this unfortunately causes a failed [build](https://travis-ci.org/masschallenge/django-accelerator/jobs/536767183) any objections to reverting the logic?

accelerator/models/profile_query_set.py
```diff
@@ -56,22 +53,13 @@ def get(self, *args, **kwargs):
...
-                # Should be able to remove this when we define a custom User
-                self.user = User.objects.filter(
-                    email=kwargs[EMAIL_KEY]).first()
+            self.user = User.objects.get(email=kwargs[EMAIL_KEY])
+
         profile_manager = self._profile_manager_by_user_type()
 
         profile = self._profile_for_inferred_profile_type(profile_manager)
         if profile:
             return profile
```
jonkiparsky: If no profile, implicitly return `None` and hope for the best?

accelerator/migrations/0054_revoke_delete_user_perms_from_non_superuser.py
jonkiparsky: Failed build because there's no such permission sometimes? 
Sounds simple enough to catch. 
The only reason I'd be concerned about reverting is I want to make sure our migrations are as zippy as possible, and looping over a large number of objects will add some friction

accelerator/tests/test_get_profile.py
jonkiparsky: Okay, so what happens if we call get_profile and there's no profile? 

This is not the usual flow for profile creation, it's a backstop for a bug that we've seen in the past. That's why it's logged, because we would like to know when it happens (and as far as I know it's been extremely rare, so I'm a bit surprised to find that we think it's causing bugs)

VincentHokie: I put back the logging for a situation where this happens

accelerator/migrations/0054_revoke_delete_user_perms_from_non_superuser.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.contrib.auth.models import Permission
+
+
+def revoke_delete_user_perms_from_non_superusers(apps, schema_editor):
```
jonkiparsky: function name is no longer accurate :)

VincentHokie: tru, updated.

mkiterian:
Taking QA

VincentHokie:
migration updated.

PR #196, opened 2019-06-14
AC-6775  Mentors and Goals unavailable after Program Ends

geofrocker:
#### Changes introduced in [AC-6775](https://masschallenge.atlassian.net/browse/AC-6775)
- Modify the `has_user_role_base` function to checks if a user is a finalist in a program that has either active or has ended
### How to test
- Follow the test instructions on the ``accelerate` [repo](https://github.com/masschallenge/accelerate/pull/2204)

### DIFF
--- a/accelerator_abstract/models/base_user_role.py
+++ b/accelerator_abstract/models/base_user_role.py
@@ -59,8 +59,7 @@ def has_user_role_base(
     if not inactive_programs:
         filter = filter.filter(program_role__program__program_status__in=[
             "active", "upcoming"])
-        return filter.exists()
-    if inactive_programs and active_or_ended_programs:
+    elif active_or_ended_programs:
         filter = filter.filter(program_role__program__program_status__in=[
             "active", "ended"])
     return filter.exists()
--- a/accelerator_abstract/models/base_user_role.py
+++ b/accelerator_abstract/models/base_user_role.py
@@ -57,11 +57,11 @@ def has_user_role_base(
     if program:
         filter = filter.filter(program_role__program=program)
     if not inactive_programs:
-        filter=filter.filter(program_role__program__program_status__in=[
+        filter = filter.filter(program_role__program__program_status__in=[
             "active", "upcoming"])
         return filter.exists()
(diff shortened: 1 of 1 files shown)


VincentHokie:
:face-palm: this function has sooo many areas of improvement ...it would be so much cleaner if it took a list of roles to check for. but this is not your code and changing that now would have you touching dozens of files. I'll write a ticket for it so that it can be tackled as it's own monster.

accelerator_abstract/models/base_user_role.py
```diff
@@ -50,14 +50,19 @@ class Meta(AcceleratorModel.Meta):
...
+        inactive_programs=False, active_or_ended_programs=False):
     filter = user.programrolegrant_set.filter(
         program_role__user_role__name=user_role_name)
     if program:
         filter = filter.filter(program_role__program=program)
     if not inactive_programs:
         filter = filter.filter(program_role__program__program_status__in=[
             "active", "upcoming"])
+        return filter.exists()
+    if inactive_programs and active_or_ended_programs:
```
jonkiparsky: This is somewhat confusing, but that's probably because it was a little confusing to begin with. 
The problem is probably that the semantics of the arguments is kind of weird. `inactive_programs` filters things out when it's False, but `active_or_ended_programs` filters things out when it's True. Shouldn't matter, but it does make it hard to read. 

One thing that would maybe make it a little bit less confusing for me would be to eliminate the `if inactive_programs` from line 63. We've already handled that condition and returned if it's False, so checking it again just makes me wonder what I've missed. 
Alternatively, you could do it like so:

```
    if not inactive_programs:
        filter = filter.filter(program_role__program__program_status__in=[
            "active", "upcoming"])
    elif active_or_ended_programs:
        filter = filter.filter(program_role__program__program_status__in=[
            "active", "ended"])
    return filter.exists()
```

geofrocker: Fixed. Thanks

VincentHokie:
will do a final QA pass once code has been checked to a ✅ state

PR #203, opened 2019-06-21
[AC-6800] Avoid routing to judge pages if not in active judging round

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-6800

Note: unusually, the fix for this is on django-accelerator and the test is on accelerate. Should be obvious why once you look at both. 
Sibling PR: https://github.com/masschallenge/accelerate/pull/2223

Testing notes:
You'll need to find or construct a user with:
- mentor program role grant and judge program role grant in the same program
- no program role grants in programs ending after that program

`40355-user@example.com` is one such user. 

To see the error state, sign in as this user on development. You will be redirected to `/panels` but because the judging round is inactive, you will then be re-redirected to the expert profile page. 

To see the fixed state, check out branch AC-6800 and navigate to http://localhost:8181/ 
This will take you to the mentor dashboard for Switzerland, which is what the ticket asks for. 

QA will be a little persnickety here, as this work depends on an interaction between code state (specifically the path starting at `calc_landing_page`) and db state (including the landing_page of the program roles assigned to the judge and the existence or non- of the related fluent pages)

### DIFF
--- a/accelerator_abstract/models/base_core_profile.py
+++ b/accelerator_abstract/models/base_core_profile.py
@@ -37,6 +37,8 @@ UI_GENDER_CHOICES = (
     GENDER_PREFER_NOT_TO_STATE_CHOICE,
     GENDER_OTHER_CHOICE,
 )
+JUDGE_FIELDS_TO_LABELS = {'desired_judge_label': 'Desired Judge',
+                          'confirmed_judge_label': 'Judge'}
 
 
 @python_2_unicode_compatible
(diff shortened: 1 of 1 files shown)


jonkiparsky:
@mkiterian can you signal whether you're taking responsibility for QA or code review? Thanks!

mkiterian:
taking QA

accelerator_abstract/models/base_core_profile.py
```diff
@@ -37,6 +37,8 @@
     GENDER_PREFER_NOT_TO_STATE_CHOICE,
     GENDER_OTHER_CHOICE,
 )
+JUDGE_FIELDS_TO_LABELS = {'desired_judge_label': 'Desired Judge',
+                          'confirmed_judge_label': 'Judge'}
```
VincentHokie: the only thing I'm curious about is whether these are the only roles that could lead to this behaviour. it seems to me that as long as a user is not a startup team member they could be affected by a similar problem. no?

as far as the ticket definition, this is out of scope I believe. but wanted to know what your views are on this

jonkiparsky: These are the only roles affected by a judging round's state...

mkiterian:
Passed QA https://github.com/masschallenge/accelerate/pull/2223#issuecomment-505437763. There's a blocking comment on the related ticket

PR #217, opened 2019-07-08
[AC-6789] EIT only finalists should not be in the Alumni program

mkiterian:
### Changes introduced in [AC-6789](https://masschallenge.atlassian.net/browse/AC-6789):
- delete global alumni startup statuses for eit only startups
- delete global alumni program role grants for eit only startup team members

### How to test
I've observed the admin views for these pages to be very slow. Shell commands have been included for convenience
**On development** 
**EIT only finallst**
- Visit http://localhost:8181/admin/simpleuser/user/57594/change/?_changelist_filters=q%3D57086 and notice that the user has both program role grants for `alum` and `eit finalist`
```
from mc.models import ProgramRoleGrant
ProgramRoleGrant.objects.filter(person_id=57594)
```
- Visit http://localhost:8181/admin/mc/startup/21869/change/ and notice the startup has startup statuses for `alum` and `eit finalist`
```
StartupStatus.objects.filter(startup_id=21869)
```

**On AC-6789**
- While checked out on this branch run `make migrate` on impact
**EIT only finallst**
- Visit http://localhost:8181/admin/simpleuser/user/57594/change/?_changelist_filters=q%3D57086 and notice that the user no longer hasa program role grant for `alum` 
```
ProgramRoleGrant.objects.filter(person_id=57594)
```
- Visit http://localhost:8181/admin/mc/startup/21869/change/ and notice the startup no longer has the startup status for `alum`
```
StartupStatus.objects.filter(startup_id=21869)
```


mkiterian:
Will wait to hear back from @EcoLek regarding startups that are in both the EIT program and the Switzerland program

mkiterian:
The resolution is to remove all eit finalists from the global alumni program since those that are other programs are only those that are also in the Switzerland program

VincentHokie:
will take code

geofrocker:
Doing QA

accelerator/migrations/0062_remove_eit_finalists_from_alum.py
```diff
@@ -0,0 +1,56 @@
...
+
+    alum_program_family = ProgramFamily.objects.filter(
+        name="Global Alumni Program").first()
+    eit_program_family = ProgramFamily.objects.filter(
+        name="EIT Food Accelerator").first()
+
+    eit_finalist_startups = StartupStatus.objects.filter(
+        program_startup_status__program__program_family=eit_program_family,
+        program_startup_status__startup_role__name=BaseStartupRole.FINALIST
+        ).values_list("startup", flat=True)
```
VincentHokie: `.distinct()` perhaps? I tried it on staging and there seems to be only one duplicate but it may be a different story on prod

VincentHokie: not a blocker though

PR #223, opened 2019-07-15
AC-6873 :: Mentors in a single program who have participated in previous years should not see program labels

freshprincekla:
**Changes introduced in [AC-6873](https://masschallenge.atlassian.net/browse/AC-6873)**
- add confirmed mentor program families to the base core profile, which returns even programs that ended.

**How to test:**
- Check out this branch 
- Masquerade as [user](http://localhost:8181/admin/simpleuser/user/844/masquerade/)
- Notice that they don't have a label on the `Mentor Dashboard` sub nav item despite having participated in previous years.

- Masquerade as [user](http://localhost:8181/admin/simpleuser/user/386/masquerade/)
- Notice that they are two `Mentor Dashboard` sub nav items with labels because they belong to different program families

**How to reproduce:**
- Masquerade as [user](http://localhost:8181/admin/simpleuser/user/844/masquerade/)
- Notice that the user  has a label on the `Mentor Dashboard` sub nav item because of having participated in previous year programs.

### DIFF
--- a/accelerator/tests/test_core_profile.py
+++ b/accelerator/tests/test_core_profile.py
@@ -225,3 +225,9 @@ class TestCoreProfile(TestCase):
         expert = ExpertFactory()
         attr = hasattr(expert.get_profile(), "confirmed_mentor_programs")
         self.assertTrue(attr)
+
+    def test_expert_profile_has_cconfirmed_memtor_program_families_all_prop(self):
+        expert = ExpertFactory()
+        attr = hasattr(expert.get_profile(),
+                       "confirmed_memtor_program_families_all")
+        self.assertTrue(attr)
--- a/accelerator/tests/test_core_profile.py
+++ b/accelerator/tests/test_core_profile.py
@@ -16,12 +16,16 @@ from accelerator.tests.factories import (
     StartupFactory,
     UserRoleFactory,
     ProgramFamilyFactory,
-    PartnerTeamMemberFactory
+    PartnerTeamMemberFactory,
 )
 from accelerator_abstract.models import (
-    BaseUserRole
+    BaseUserRole,
+    ENDED_PROGRAM_STATUS,
 )
-from accelerator.tests.contexts import StartupTeamMemberContext
+from accelerator.tests.contexts import (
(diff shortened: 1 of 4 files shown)


accelerator/tests/test_core_profile.py
```diff
@@ -225,3 +225,10 @@ def test_expert_profile_has_confirmed_mentor_programs_prop(self):
         expert = ExpertFactory()
         attr = hasattr(expert.get_profile(), "confirmed_mentor_programs")
         self.assertTrue(attr)
+
+    def test_expert_profile_has_confirmed_mentor_program_families_all_prop(
```
VincentHokie: I would prefer something along the lines of
 **untested**
```
def test_user_profile_confirmed_mentor_program_families_method(self):
program1 = ProgramFactory()
program2 = ProgramFactory()
user = EntrepreneurFactory()
context = UserRoleContext(
    UserRole.MENTOR,
    user=user,
    program=program1)
 UserRoleContext(
    UserRole.MENTOR,
    user=user,
    program=program2)
families = user.get_profile().confirmed_memtor_program_families_all()
self.assertTrue(program1.program_family in families)
self.assertTrue(program2.program_family in families)
```

VincentHokie:
I'm on code

accelerator/tests/test_core_profile.py
```diff
@@ -225,3 +229,25 @@ def test_expert_profile_has_confirmed_mentor_programs_prop(self):
...
+        UserRoleContext(
+            BaseUserRole.MENTOR,
+            user=user,
+            program=active_program)
+        UserRoleContext(
+            BaseUserRole.MENTOR,
+            user=user,
+            program=inactive_program)
+        families = user.get_profile().confirmed_memtor_program_families_all()
+        self.assertEqual(len(families), 2)
```
VincentHokie: I think we should add the assertion you had here, it was failing coz you were checking for an object in a list of strings, in addition to what you have now
```
self.assertTrue(active_program.program_family.name in families)
self.assertTrue(inactive_program.program_family.name in families)
```
should make this test more concrete

PR #230, opened 2019-07-29
[AC-6924] 'office_hour_is_in_past' utility function raises exception when 'start_date_time' is None

mkiterian:
### Changes introduced in [AC-6924](https://masschallenge.atlassian.net/browse/AC-6924):
- Make start_date_time and end_date_time fields non nullable

### How to test
- This change can't be tested directly


kitaka:
Taking code review

kitaka:
@mkiterian you should add a test

mkiterian:
@kitaka we do not have an interface that would allow the creation of an office hour without a providing a start/end datetime. Testing this through model creation is like testing the django framework

kitaka:
@mkiterian my thinking is that if an exception is being thrown on production, it should be reproducible with a test.

kitaka:
@mkiterian as discussed on the call, the PR title was a bit misleading and its what inspired my request for a test. Given that this is a break away from the original ticket, it might have been better served as a micro ticket.

geofrocker:
QA looks good. The following has been tested
- Migrations running successfully
- Office hour operations for finalist, mentor and staff

EcoLek:
Is there a reason this hasn't been merged? Two approvals @kitaka @geofrocker

geofrocker:
Not really

PR #232, opened 2019-08-01
[AC-6727] DUE End of August - Legal Checks are capable of being shown to particular user type

kitaka:
**Changes Introduced in [AC-6727](https://masschallenge.atlassian.net/browse/AC-6727):**
- Added ability for Legal checks to be shown to Experts profile types and/or Entrepreneurs profile

**Test:**

_checkout this branch_
- Run migrations `make migrate`
- Create Legal Check enabled for entrepreneurs only and another enabled for experts only (here: http://localhost:8181/admin/mc/legalcheck/add/)
- Visit the home page http://localhost:8181/ and see only the entrepreneur check appears
- Approve the check and notice you aren't prompted again

Related PR: https://github.com/masschallenge/accelerate/pull/2293


mkiterian:
Taking code

accelerator_abstract/models/base_legal_check.py
```diff
@@ -21,6 +21,13 @@ class BaseLegalCheck(AcceleratorModel):
                             help_text='Internal name for this check.')
     description = models.TextField(
         help_text='Text displayed next to checkbox. Use HTML for links.')
+    is_enabled_for_experts = models.BooleanField(
+        default=True,
```
mkiterian: why default to True?

kitaka: @mkiterian that was to ensure that the old Legal Checks weren't affected by the new changes.

PR #234, opened 2019-08-02
AC-6901 Display title and company on the cards for experts in the people directory

geofrocker:
#### Changes introduced in [AC-6901](https://masschallenge.atlassian.net/browse/AC-6901)
- Add logic for using expert profiles incase the user profile does not have the required values
##### On localhost
- Checkout this branch on `django-accelerator`.
##### On `accelerate`
- update the setting file with ALGOLIA KEYS from 1password and set ALGOLIA_ENABLED = True
- Run `make bash-shell`
- Re-index the user model `bin/django algolia_reindex --model User`(This will take some time)
- Run impact and navigate to the [people directory](http://localhost:1234/people)
- Notice that most of the cards have the test (title and company) data on them
##### On test6
- Navigate to the people [directory](https://test6.masschallenge.org/people/)
- Search for `Jeanne Zalesky` or follow this [link](https://test6.masschallenge.org/people/?query=Jeanne%20Zalesky)
- Notice that there is a user title and startup name. This was not the case previously. Double check with the ticket description.


simpleuser/models.py
```diff
@@ -116,7 +121,12 @@ def type(self):
         return self._get_profile().user_type
 
     def startup_name(self):
-        return self.startup.name if self._get_startup() else None
+        if self._get_startup():
```
jonkiparsky: Note that if an expert is associated with a startup, that will be what is returned rather than their company name. We should check that this is the desired behavior, and if it is we should write a test to explicitly see that happening.

simpleuser/tests/test_models.py
```diff
@@ -84,10 +84,17 @@ def test_current_program(self):
         user.program()
         startup.current_program.assert_any_call()
 
-    def test_expert_startup_name_is_none(self):
-        user = ExpertFactory()
-        user = User.objects.get(pk=user.pk)
-        self.assertEqual(user.startup_name(), None)
+    def test_expert_startup_name_is_not_none(self):
```
jonkiparsky: We should have a test for the case where an expert has a startup. As noted above, we should explicitly confirm with product whether in that case we want to display profile.company or one of their startups. (as written, we're doing the latter)

jonkiparsky: Sorry, I was not clear. I want *both* tests - one for with an explicit startup, one without. 
Two logical flows, two tests.

geofrocker: Waiting to sync with Product

jonkiparsky: @geofrocker I've talked with @hfreund and @EcoLek about this, and both are agreed that we should use the information from the expert's profile first. Presumably, if that field is empty - 18 such cases exist now - we should look for a startup and use that.

geofrocker: Thank you. It makes more sense

simpleuser/models.py
```diff
@@ -116,6 +120,9 @@ def type(self):
         return self._get_profile().user_type
 
     def startup_name(self):
+        profile = self._get_profile()
+        if self._is_expert():
+            return profile.company
```
jonkiparsky: If the company field is blank, we should probably try their startup. 

This will require a bit of refactor, since in that case we'd want the title and company name to come from the same place, so user_title and startup_name will have to be unified. 

From talking with @EcoLek he's fine with taking the time to get this sorted out. 

The cases are:
- expert has title and/or company listed on profile: use info from profile
- else: try a startup (and use title and company from that startup)
- if no startup exists, list nothing. 

(in future, we may give users a bit more control over what gets displayed, but not today)

geofrocker: Noted

simpleuser/models.py
```diff
@@ -121,20 +121,26 @@ def startup_name(self):
...
+        title = self.title if self._get_title() else ""
+        company = self.startup.name if self._get_startup() else None
         return {
             "title": title,
             "company": company
         }
+    def _has_expert_details(self):
+        if self._is_expert():
+            profile = self._get_profile()
+            title = hasattr(profile, "tile") and profile.title
```
jonkiparsky: Why `hasattr` here? An expert profile always has the `title` and `company` attributes, since they're defined on the model. The question is whether the attributes have non-empty values, so this can be written as 
```
    def _has_expert_details(self):
        if self._is_expert():
            profile = self._get_profile()
            return True if profile.title or profile.company
```

simpleuser/models.py
```diff
@@ -120,10 +118,23 @@ def type(self):
...
 
     def startup_name(self):
+        return self._get_title_and_company()['company']
+
+    def _get_title_and_company(self):
         profile = self._get_profile()
-        if self._is_expert():
-            return profile.company
-        return self.startup.name if self._get_startup() else None
+        title = profile.title if profile.title else "",
```
jonkiparsky: since `profile.title` is a string, you could just have `title = profile.title` (and likewise `company = profile.company`)

simpleuser/models.py
```diff
@@ -120,10 +118,23 @@ def type(self):
...
+        company = profile.company if profile.company else None
```
jonkiparsky: `else None`? I would think this should be a string, see comment above.

simpleuser/models.py
```diff
@@ -116,7 +118,30 @@ def type(self):
         return self._get_profile().user_type
 
     def startup_name(self):
-        return self.startup.name if self._get_startup() else None
+        return self._get_title_and_company()['company']
+
+    def _get_title_and_company(self):
+        if self._is_expert() and self._has_expert_details():
+            profile = self._get_profile()
+            title = profile.title if profile.title else "",
```
jonkiparsky: since `profile.title` is a string, you could just have `title = profile.title` (and likewise `company = profile.company`)

simpleuser/models.py
```diff
@@ -116,7 +118,30 @@ def type(self):
...
+            company = profile.company if profile.company else None
```
jonkiparsky: `else None`? I would think this should be a string, see comment above.

simpleuser/tests/test_models.py
```diff
@@ -84,10 +85,35 @@ def test_current_program(self):
...
+
+    def test_user_without_a_startup(self):
+        user = UserFactory()
+        self.assertEquals(user.startup_name(), None)
+
+    def test_expert_has_a_user_title(self):
+        expert = ExpertFactory()
+        title = expert.user_title()
+        user = User.objects.get(pk=expert.pk)
+        self.assertEqual(user.user_title(), title)
```
jonkiparsky: This assertion amounts to `user.user_title() == user.user_title()`. You could do `title = expert.get_profile().title` instead

geofrocker: Thanks. same for expert startup

PR #236, opened 2019-08-08
[AC-6946] Enable smart locations on Accelerate Program families

kitaka:
**Changes Introduced in [AC-6946](https://masschallenge.atlassian.net/browse/AC-6946):**
- Added Location and ProgramFamilyLocation models
- Added Admin interfaces for the models above

**Test:**

_checkout this branch_
- Visit http://localhost:8181/admin/ and notice you can administer `Location`, `Program Family Location` and you can also add `Program Family Location` as a fieldset on `Program Family`

Sibling PR: https://github.com/masschallenge/accelerate/pull/2300


kaymation:
"i'm trying ringo, I'm trying real hard... to be the shepherd"

accelerator/tests/factories/location_factory.py
```diff
@@ -0,0 +1,23 @@
...
+        model = Location
+
+    name = Sequence(lambda n: "Location {0}".format(n))
+    city = "Boston"
+    state = "Massachusetts"
+    country = "United States"
+    postcode = "02210"
+    latitude = "100"
+    longitude = "200"
+    timezone = "America/New_York"
```
kaymation: hmm. food for thought. Are there any disadvantages to having our factory location default to values we'd expect to see in prod? we seeeem to avoid that elsewhere, i think because it might encourage dogmatic implicitness as opposed to clear explicitness in written tests. very much a nonblocking concern

kaymation:
Should we consider at least having the display say "state/province" or something else?
For the actual field name i think 'state' is fine, but it's possible users in other countries may have provinces, or even a different term.
Googling around shows that Israel has districts, Switzerland has Cantons.
Might not be important, but something to consider

accelerator_abstract/models/base_location.py
```diff
@@ -0,0 +1,32 @@
...
+)
+
+from accelerator_abstract.models.accelerator_model import AcceleratorModel
+
+
+class BaseLocation(AcceleratorModel):
+    name = CharField(max_length=100, blank=False)
+    city = CharField(max_length=100, blank=True, default='')
+    state = CharField(max_length=100, blank=True, default='')
+    country = CharField(max_length=100, blank=True, default='')
```
kaymation: could this be a choice field? Even if we dont want to limit to the countries of our PFs, there are only 195 countries and major regime changes dont happen that often

accelerator_abstract/models/base_location.py
```diff
@@ -0,0 +1,32 @@
...
+    postcode = CharField(max_length=20, blank=True, default='')
+    latitude = CharField(max_length=100, blank=True, default='')
+    longitude = CharField(max_length=100, blank=True, default='')
+    timezone = CharField(
+        max_length=35,
+        blank=False,
+        default="UTC",
+        help_text=("Timezone name from Olson Timezone database "
+                   "(https://en.wikipedia.org/wiki/Tz_database)")
+    )
```
kaymation: i get that there are technically a loooot of time zones, but this being selectable with ~24 opts would also be cool

accelerator_abstract/models/base_program_family_location.py
```diff
@@ -0,0 +1,29 @@
...
+    location = models.ForeignKey(
+        swapper.get_model_name(AcceleratorModel.Meta.app_label,
+                               "Location"))
+    primary = models.BooleanField()
+
+    class Meta(AcceleratorModel.Meta):
+        db_table = '{}_programfamilylocation'.format(
+            AcceleratorModel.Meta.app_label)
+        abstract = True
+        unique_together = ('program_family', 'location')
```
kaymation: nothing is stopping you from adding more than one primary location, verified when QAing

kitaka: @kaymation  I considered adding primary to the composite key, however, it seems it would also limit duplicate non-primary locations. What do you think?

accelerator_abstract/models/base_location.py
kitaka: @kaymation I based it off the timezone field under ProgramFamily. I didn't want to reinvent the wheel. Not sure about the tradeoffs between a selectable and non-selectable. @jonkiparsky, @hfreund and @EcoLek can add their thoughts to it given they approved the current non-selectable field.
https://github.com/masschallenge/django-accelerator/blob/312dc72df7aad67a3e059f16efb59288447b9fc8/accelerator_abstract/models/base_program_family.py#L61

accelerator_abstract/models/base_location.py
kitaka: @hfreund @EcoLek any thoughts on this? The ticket wasn't specific.

accelerator/tests/factories/location_factory.py
kitaka: @jonkiparsky What's your opinion on this one?

kitaka:
> Should we consider at least having the display say "state/province" or something else?
> For the actual field name i think 'state' is fine, but it's possible users in other countries may have provinces, or even a different term.
> Googling around shows that Israel has districts, Switzerland has Cantons.
> Might not be important, but something to consider

@kaymation Algolia uses `administrative` to represent the same field. I wonder what @hfreund's thoughts are on it.

accelerator/tests/factories/location_factory.py
jonkiparsky: I'm not too concerned about this one, but sure, it might make sense to pick a location that is not associated with one of our current programs, to ensure that we're not writing in expectations like "America/New_York" into the code.

accelerator_abstract/models/base_location.py
jonkiparsky: I think it's probably premature optimization. We want to provide a proper staff setup page for this, and we should focus our UI efforts there rather than on trying to shine up admin pages that we don't want to keep. 

There is a good library for location selection that we use in the registration process now, we can use that when we do the UI phase on this.

hfreund: After discussing with Felix, I am comfortable leaving this as an open text field since each country is written only a few times and the likelihood of it being misspelled or written differently between two program families is low given that only a few high-level staff operators will manage this information. Felix was smart to point out that the value gained by using a dropdown is not worth the effort to implement - I agree.

accelerator_abstract/models/base_location.py
hfreund: My intuition is to use a pre-defined list of time zones rather than ask users to know which one to input. However, I understand Jon has good context around how to implement this so I would like to hear his comments.

If this remains as an open text field I suggest two changes to the help text:
1. Indicate the expected input value so that users who are unfamiliar with timezones can be confident they are writing the correct value
2. Include a link to the actual list of Olsen time zones - the current Wikipedia article linked in the help text provides examples and information about time zones but does not provide an exhaustive list of all timezones.

jonkiparsky:
@kaymation We use "State/Region/Province" on the startup profile, we should probably use the same on any user-facing interface but not worry too much about it now.

accelerator_abstract/models/base_program_family_location.py
kaymation: why would someone wnt to duplicate non-primaries anyways?

jonkiparsky: good point

kitaka: @kaymation the composite key would be `('program_family', 'primary')` 
Given that a program family can have more than one non-primary location, the constraint might make it impossible. Could I be missing something?

jonkiparsky: @kitaka the unique_together would be `('program_family', 'location', 'primary')`
Which would actually do what we want, except that it would be possible to have the Boston Program Family have Boston as a Location twice, once as primary and once as non-primary.

kitaka: @jonkiparsky I don't know how it would solve the main issue of ensuring the a Program Family has a single primary Location. Since the locations will be different, they will be unique even though the program family is the same.

jonkiparsky: Doh. Yes, you're right. I don't know where my brain was on Friday, please ignore.

kaymation: I think the only away around this is a presave hook

kitaka: @kaymation that has been implemented on the Sibling PR

PR #241, opened 2019-08-20
[AC-5845] Too-long topics in office-hours reservation request should not throw a 500

freshprincekla:
[JIRA](https://masschallenge.atlassian.net/browse/AC-5845)

We are getting FLOODED with errors from folks who try to submit an office hours topic with length > 500 characters.

We would love for this form to be a semantified one with modern validation, but removing our legacy code and replacing it with a contemporary form is many points of work, not within the scope of this urgent fix. 

TO TEST - 
* Sign in as a user in an open program
* go the Office Hours page
* try to reserve an hour and put a topic that is >500 characters
* No error will occur

[This ticket](https://masschallenge.atlassian.net/browse/AC-6986) covers a long term solution, which requires UX and estimation


geofrocker:
Taking the code

mkiterian:
Taking QA

PR #251, opened 2019-09-06
[AC-6763-b] Added location ForeignKey Field.

kitaka:
**Changes Introduced in [AC-6763](https://masschallenge.atlassian.net/browse/AC-6763):**
- Added location ForeignKey Field to MentorProgramOfficeHour model

**Test:**

- Passing tests are sufficient as only a field is added.


shakirandagire:
Doing QA for the task

PR #254, opened 2019-09-11
[AC-7049]: Program cycles associated with a program can have no application type

shakirandagire:
**Changes introduced in [AC-7049](https://masschallenge.atlassian.net/browse/AC-7049):**
- Added the conditions for the open applications must have a default application type and default application may not be removed from a program cycle when that program cycle is associated with a program and applications are open.

**How to test:**
**On development:**
- [Create a program cycle](http://localhost:8181/admin/mc/programcycle/add/) using the "Steps to repeat" from the ticket.
Note that the program cycle is created properly.
- [Create a program](http://localhost:8181/admin/mc/program/add/) using "Steps to repeat" from the ticket.
Note that the program is not saved and an error is triggered.
This error vanishes when the associated program cycle is given a default application type.
Note that the program is now saved.

**On this branch:**

- Follow the steps above to create the program cycle and program.
- [Create a program cycle](http://localhost:8181/admin/mc/programcycle/add/) using the "Steps to repeat" from the ticket
- If the application open checkbox is checked and the default application type is not selected an error is triggered.
- If the application open checkbox isnot checked and the default application type is not selected and the program cycle is saved.


shakirandagire:
> QA issue:
> I tested the case of an existing program cycle already associated with several programs (http://localhost:8181/admin/mc/programcycle/35/change/).
> I turned off the "applications open" switch and set the Default Application Type to the empty value, and was able to save without seeing an error message.
> @shakirandagire is this expected?
> The ticket says "However, it appears to be possible to create an application type, associate that with a program cycle and that with a program and then remove the application from the program cycle. This should also fail validation".

@jonkiparsky 
That is the expected behaviour when the program cycle doesn't have open applications.
The ticket's done specifies that the errors are triggered when the program cycle has open applications.

accelerator/tests/test_program_cycle.py
```diff
@@ -3,12 +3,19 @@
...
+
 from accelerator.tests.factories import ProgramCycleFactory
 
 
 class TestProgramCycle(TestCase):
     def test_display_name_no_short_name(self):
         cycle = ProgramCycleFactory(short_name=None)
         assert cycle.name in str(cycle)
+
+    def test_program_cycle_has_default_application_type(self):
```
pbx: Looks good, very nice update to this code. My only request - saying "requires" instead of "has" in the test name would be clearer, i.e. `test_program_cycle_requires_default_application_type`

jonkiparsky:
@shakirandagire Need to merge development into this branch

shakirandagire:
@pbx I have made the requested change.
@jonkiparsky I merged changes from development

PR #263, opened 2019-10-08
[AC-6763-c] Office hours interface use smart locations

kitaka:
**Changes Introduced in [AC-6763](https://masschallenge.atlassian.net/browse/AC-6763):**
- Office hours interface uses locations from attached program's program family

**Test:**

_checkout this branch_
- Login as `Demo Admin`
- Masquerade as the user `1032-user@example.com`
- Visit http://localhost:8181/officehours/bos/2019/
- Notice that only locations for the `Boston` program are shown

_checkout development_
- Notice that locations for all programs are shown

Sibling PR: https://github.com/masschallenge/accelerate/pull/2396


accelerator_abstract/models/base_mentor_program_office_hour.py
```diff
@@ -52,7 +30,7 @@ class BaseMentorProgramOfficeHour(AcceleratorModel):
     start_date_time = models.DateTimeField(db_index=True)
     end_date_time = models.DateTimeField(db_index=True)
     description = models.TextField(blank=True)
-    old_location = models.CharField(max_length=50, choices=LOCATION_CHOICES)
+    old_location = models.CharField(max_length=50)
```
jonkiparsky: we should probably make old_location nullable if we're deprecating it

PR #271, opened 2019-10-29
AC-6346 - Add fields to the Expert Profile and to Expert Pool reports for Expert curation

freshprincekla:
**Changes introduced in [AC-6346](https://masschallenge.atlassian.net/browse/AC-6346)**
 - Added Mentor type and judge type to the Expert Profile model
 - Added these fields to the Mentor and judging pool.

**How to test**
- Checkout this branch and the sibling branch in accelerate.
- Run the migrations in Django-accelerate
- Login in a DemoAdmin and go to a mentor in SimpleUser admin. For example [user](http://localhost:8181/admin/simpleuser/user/27/change/)
- Fill in Mentor Type and the Judge type with the suggested input.
- Save and then navigate to The [Judge pool](http://localhost:8181/staff/judge_pool/) and the [Mentor pool](http://localhost:8181/staff/mentor_pool/) and search for the user. They will have Mentor Type and the Judge type column with the values input previously.


accelerator_abstract/models/base_expert_profile.py
```diff
@@ -51,6 +51,24 @@
 
 BIO_MAX_LENGTH = 7500
 
+MENTOR_TYPE_HELPTEXT = (
```
jonkiparsky: If these are the allowed values, shouldn't they be [choices](https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices)?

jonkiparsky: Also, see the ticket:

> Values: F, P, T, E, N, X (info after dash is informational for us only)

So we want to be writing "F", "P", etc into the db, and using the strings "Functional Expert", "Partner", etc for display values. 

Should check with product on whether they want the choice value (F, P, etc) or the display value ("Functional Expert", "Partner", etc) to show in the report. 
See https://github.com/masschallenge/django-accelerator/blob/development/accelerator_abstract/models/base_program.py#L11 or https://github.com/masschallenge/django-accelerator/blob/development/accelerator_abstract/models/base_judging_round.py#L38 for examples of how we usually handle choices on models.

freshprincekla: @jonkiparsky had talked with product, and was told to leave the field as a textbox rather than choices. Was told that the values are most likely going to change and will be coming from saleforce. Jeff said its best for them not to be choices.

accelerator_abstract/models/base_expert_profile.py
```diff
@@ -75,6 +93,16 @@ class BaseExpertProfile(BaseCoreProfile):
         verbose_name="Functional Expertise",
         related_name="experts",
         blank=True)
+    mentor_type = models.CharField(
+        max_length=64,
```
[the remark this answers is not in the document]
freshprincekla: Yeah you are right. Ooops Let me correct this

PR #275, opened 2019-11-07
[AC-7212] Office hour Ical error ('NoneType' object has no attribute 'name')

mkiterian:
### Changes introduced in [AC-7212](https://masschallenge.atlassian.net/browse/AC-7212):
- remove the old location field from office hours


jonkiparsky:
I suggest that this should merge after the sibling PR is deployed, to avoid bad db state

jonkiparsky:
Change looks good, but holding off on approve until sibling ticket is deployed

PR #276, opened 2019-11-12
[AC-7128] Make role-based landing page logic aware of state

jonkiparsky:
Added logic to parse state a little more consciously. 
confirmed judge grants should only matter for active judging rounds, desired judge grants should only matter for judging rounds which have not ended yet, mentor grants should only matter for active programs

Tests are on accelerate because that's where we actually exercise the functionality. 

[Sibling branch](https://github.com/masschallenge/accelerate/pull/2430) exists, should merge simultaneously with this branch.


jonkiparsky:
Reviewers please note: the code climate issue will require substantial refactor to resolve. I would suggest that this is not a change we want to make on this ticket.

jonkiparsky:
@shanbady I've removed the uncovered code, turns out it was no longer needed.

jonkiparsky:
@shakirandagire waiting on you for this

PR #278, opened 2019-11-14
[AC-7044] Clear all Django deprecation warnings on django-accelerator

philophilo:
### Changes introduced in: [AC-7044](https://masschallenge.atlassian.net/browse/AC-7044):
- Resolves `RemovedInDjango20Warning` warning for on_delete
- Resolves `RemovedInDjango20Warning` to use `django.urls` instead of `django.core.urlresolvers`

### How to test

- *From `development` branch*
- Modify the `Makefile` change the `test` command to use `python3 -Wall`
```
test: $(VENV)
	@$(ACTIVATE); DJANGO_SETTINGS_MODULE=settings python3 -Wall $(DJANGO_ADMIN) test $(tests)
```
- Run `make test`
- This should return `RemovedInDjango20Warning` warnings

- *Switch this branch `AC-7044`*
- Run `make test`
- This should return no warnings pertaining to `RemovedInDjango20Warning`


philophilo:
@jonkiparsky can just add the `manage.py` file to that tests are run using the `python` command. The warnings can only be passed when running `python ...` unlike `bin/django ...`. If there is an alternative, I would gladly use some help on this.

pbx:
@philophilo - `$(DJANGO_ADMIN)` in the django-accelerator Makefile is a Python script, so you can precede it with `python -Wa`, like is done here with `bin/django`: https://github.com/masschallenge/accelerate/pull/2432/files#diff-b67911656ef5d18c4ae36cb6741b7965R176

philophilo:
@pbx @jonkiparsky  Since we do not have a container isolating this repo locally, from the `Makefile` we create the virtual environment using the python version installed on the host machine. Just highlighting an inconsistency.
https://github.com/masschallenge/django-accelerator/blob/9d9f0c8e200ab0143fe02302d655b7a5fc26d5e1/Makefile#L99-L112

pbx:
Good catch @philophilo , I'd say that's worth at ticket. Ideally the host environment should not be a factor in the way the software runs.

pbx:
I told Phil I'll take care of the CodeClimate "duplicate code" warnings.

PR #281, opened 2019-11-19
[AC-7263] Fix for redirect issue introduced in AC-7128

jonkiparsky:
see ticket for details. 

https://masschallenge.atlassian.net/browse/AC-7263


jonkiparsky:
oops!

PR #286, opened 2019-11-21
[AC-7272] Updated copy from a to a(n) for expert category verbose name

MCGigi:


jonkiparsky:
Added migration. Should be good now.

PR #288, opened 2019-11-27
removing max version constraint for django

shanbady:
This pr removes the max django version constraint of 2.0 from django-accelerator. This is so that we do not have to monkey patch the dockerfile in the django 2 branch as we continue work there. We should be able to safely merge this into development.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ with open(os.path.join(os.path.dirname(__file__), 'README.md')) as readme:
 # allow setup.py to be run from any path
 os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
 
-DJANGO_SPEC = ">=1.8"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -11,7 +11,7 @@
 # allow setup.py to be run from any path
 os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
 
-DJANGO_SPEC = ">=1.8,<2.0"
+DJANGO_SPEC = ">=1.8"
```
pbx: Since we're aiming for 2.2 LTS, let's make this `>=1.11,<=2.3`

shanbady: fixed

pbx: Oops, sorry @shanbady , I meant `>=1.11,<2.3` -- we don't want 2.3!

PR #289, opened 2019-11-28
[AC-6592] Update context to allow a test to work right

jonkiparsky:
This fixes a simple oversight in this context. Should merge before the sibling PR.


geofrocker:
Code looks, I need clarification on the last bit of the QA
- `Request applications` not firing any action

PR #290, opened 2019-12-02
[AC-7030] Timebox - Update dependencies of django-accelerator for Django 2 compatibility

philophilo:
### Changes introduced in: [AC-7030](https://masschallenge.atlassian.net/browse/AC-7030):
- Bumps up django version

### How to test
- Travis build with Django 2.2 does not fail

local testing
- run `make setup`, after setup
- run `make run command=<any make command in the Makefile>` like `make run command=test`
- run `make pull-down` to stop the container
- No warnings while running tests


base.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    chown -R accelerate_user /var/lib/ && \
+    sed -i  '/requiretty/s/^/#/'  /etc/sudoers && \
+    echo "accelerate_user ALL=(ALL)       NOPASSWD: ALL" >> /etc/sudoers && \
+    echo "accelerate_user -  nofile 65535" >> /etc/security/limits.conf && \
+    apt-fast -y install python-setuptools python-dev emacs24 libjpeg8-dev && \
+    sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.4 1 && \
+    sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.6 2 && \
+    sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.6 3
+
+RUN wget https://www3.sqlite.org/cgi/src/tarball/fc82b73eaa/SQLite-fc82b73eaa.tar.gz && \
```
jonkiparsky: Drive-by: this looks very hard-coded and fragile and hard to understand. I'd be a little concerned about maintainability here, is there any way to make it more communicative and robust?

scripts/install_sqlite.sh
```diff
@@ -0,0 +1,21 @@
...
+
+remove_older_version(){
+    echo "======= sqlite version current ======="
+    sqlite3 --version
+    sudo apt-get purge --auto-remove sqlite3
+}
+
+install_sqlite(){
+    # install sqlite 3.29.0
+    wget https://www3.sqlite.org/cgi/src/tarball/fc82b73eaa/SQLite-fc82b73eaa.tar.gz
```
jonkiparsky: Drive-by: similar concern, this looks like a a potential maintainability snag.

base.Dockerfile
philophilo: Aaah I see, thanks, let me find a better way to installation

pbx:
(Additional note -- the work in this PR takes care of [AC-7033](https://masschallenge.atlassian.net/browse/AC-7033), so that's nice!)

philophilo:
@pbx In this `base.Dockerfile`, I have removed some of the unnecessary packages like MySQL since `django-accelerator` uses sqlite. But indeed it may still need more cleanup.

pbx:
@philophilo - I'm not as concerned about cleanup as I am about maintenance -- if the two Dockerfiles diverge (e.g. if we upgrade to Python 3.8 in one and not in the other) and we start seeing inconsistent behavior, we may lose time trying to figure out the cause.

If they have to be like this, so be it, but if there's a way for them to share a common base that could address this concern.

pbx:
My takeaway from this PR is:
* There are no changes needed to django-accelerator's dependencies for compatibility with Django 2
* The work in this branch lets us confirm that (inasmuch as we trust our test suite)
* This branch does not need to be merged in order for the ticket to be marked Done (and might in fact be better used in [AC-7258](https://masschallenge.atlassian.net/browse/AC-7258).

pbx:
Shankar has requested we close this PR for now. 

Re the ticket (distinct from the PR), I'd argue that it can be marked Done based on the successful test-run under Django 2 (which this ticket's changes make easy, thanks!). Need at least one other person to agree though :)

PR #291, opened 2019-12-10
[AC-7033] Update django-accelerator Django version to 2.2

pbx:
## [AC-7033](https://masschallenge.atlassian.net/browse/AC-7033)

**Testing**: 
* In your local django-accelerator clone on this branch, run `rm -r venv && make test`
  * Confirm that Django 2 was in fact used: `. venv/bin/activate && python -c "import django; print(django.VERSION)"`
* Run `tox`, which will run the test suite under both Python 3.6 and Python 3.7.
  * Confirm that Django 2 was in fact used: Look for output like `py37 installed:...Django==2.2.8`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: xenial
 language: python
 python:
   - "3.6"
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ help:
 	@echo
 
 
-DJANGO_VERSION = 1.11.18
+DJANGO_VERSION = 2.2.8
 VENV = venv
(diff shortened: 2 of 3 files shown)


pbx:
I'm adding the **WIP** label because of this Travis error, which I don't know the cause of: 
```
File "/home/travis/build/masschallenge/django-accelerator/venv/lib/python3.6/site-packages/django/db/backends/sqlite3/base.py", line 63, in check_sqlite_version
    raise ImproperlyConfigured('SQLite 3.8.3 or later is required (found %s).' % Database.sqlite_version)
django.core.exceptions.ImproperlyConfigured: SQLite 3.8.3 or later is required (found 3.8.2).
```

PR #293, opened 2019-12-16
AC-7258 Run django-accelerator in a docker environment

philophilo:
### Changes introduced in: [AC-7258](https://masschallenge.atlassian.net/browse/AC-7258):
- Runs django-accelerator in docker

### How to test
- run `make setup` to setup the docker environment
- run `make run command=<any make command in the Makefile>` like `make run command=test`
- run `make stop-server` to stop the container
- run `make bash-shell` to get into the running container

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,11 +84,7 @@ target_help = \
   'setup - Sets up a Docker testing environent for django_accelerator' \
   'stop-server - Stops the Docker testing environent for django_accelerator' \
   'run - runs Makefile commands inside the Docker container.' \
-  '\tSet the "command" variable to one of:' \
-  '\t [help, package, clean, code-check, coverage, coverage-run, ' \
-  '\t coverage-report, coverage-html-report, coverage-xml-report, ' \
-  '\t install, uninstall, data-migration, migrations, test, tox] '\
-  '\t E.g. `make run command=test`' \
+  '\tSet the "command" variable e.g. `make run command=test`' \
   "bash-shell - Allows getting into the django_accelerator container's terminal" \
 
 OS = $(shell uname)
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,18 @@ target_help = \
   'deploy - Deploy $$(release_name) to a $$(target).' \
   '\tValid targets include "staging" (the default), "production",' \
   '\t "test-1", and "test-2"' \
+  ' ' \
+  'install-sqlite - Installs sqlite3 on Travis from a script in' \
+  '\t scripts/run_make_commands_in_docker.sh, however it also can run in the' \
+  '\t accelerator container' \
+  'build - Builds the image for django_accelerator' \
+  'setup - Sets up a docker testing environent for django_accelerator' \
+  'pull-down - Stops the docker testing environent for django_accelerator' \
+  'run - runs [help, package, clean, code-check, coverage, coverage-run, ' \
+  'coverage-report, coverage-html-report, coverage-xml-report, ' \
+  '\t coverage-html, install, uninstall, data-migration, migrations, ' \
+  '\t test, tox, install_sqlite] set of commands from the testing docker environent' \
+  "ssh - Allows getting into the django_accelerator container's terminal" \
 
 OS = $(shell uname)
 
@@ -196,3 +208,18 @@ migrate update-schema:
 
 checkout:
 	@$(IMPACT_MAKE) $@ branch=$(branch)
+
+build:
+	@docker build -f base.Dockerfile -t accelerator_tests .
+
+setup: build
+	@docker-compose up -d
+
+pull-down:
+	@docker-compose down
+
+run:
+	@docker-compose exec accelerator scripts/run_make_commands_in_docker.sh $(command)
+
+ssh:
+	@docker-compose exec accelerator /bin/bash
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,6 @@ target_help = \
   '\tValid targets include "staging" (the default), "production",' \
   '\t "test-1", and "test-2"' \
   ' ' \
-  'install-sqlite - Installs sqlite3 on Travis from a script in' \
   '\t scripts/run_make_commands_in_docker.sh, however it also can run in the' \
   '\t accelerator container' \
   'build - Builds the Docker image for django_accelerator' \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -77,6 +77,18 @@ target_help = \
   'deploy - Deploy $$(release_name) to a $$(target).' \
   '\tValid targets include "staging" (the default), "production",' \
   '\t "test-1", and "test-2"' \
+  ' ' \
+  'install-sqlite - Installs sqlite3 on Travis from a script in' \
+  '\t scripts/run_make_commands_in_docker.sh, however it also can run in the' \
+  '\t accelerator container' \
+  'build - Builds the image for django_accelerator' \
```
pbx: Let's clearly indicate which commands work with the Docker image/container.  E.g. I'd prefer if this said "Builds the Docker image..."

Makefile
```diff
@@ -77,6 +77,18 @@ target_help = \
...
+  'setup - Sets up a docker testing environent for django_accelerator' \
```
pbx: Capitalize "Docker"

Makefile
```diff
@@ -77,6 +77,18 @@ target_help = \
...
+  'pull-down - Stops the docker testing environent for django_accelerator' \
```
pbx: Please capitalize "Docker".

Also, for consistency let's make this more like the other up/down commands in our other makefiles (which generally begin with "start" or "stop" or "shutdown") -- e.g. `shutdown-docker-container` or `shutdown-docker`

Makefile
```diff
@@ -77,6 +77,18 @@ target_help = \
...
+  'run - runs [help, package, clean, code-check, coverage, coverage-run, ' \
+  'coverage-report, coverage-html-report, coverage-xml-report, ' \
+  '\t coverage-html, install, uninstall, data-migration, migrations, ' \
+  '\t test, tox, install_sqlite] set of commands from the testing docker environent' \
```
pbx: This isn't as clear as it could be. I think the help might be a bit clearer like this: 
```
run - runs Makefile commands inside the Docker container. Set the "command" variable to one of: [help, package, clean, code-check, coverage, coverage-run,  coverage-report, coverage-html-report, coverage-xml-report, coverage-html, install, uninstall, data-migration, migrations, test, tox, install_sqlite] E.g. `make run command=test`
```

Makefile
```diff
@@ -77,6 +77,18 @@ target_help = \
...
+  "ssh - Allows getting into the django_accelerator container's terminal" \
```
pbx: Let's call this one `make bash-shell` as in the other Makefiles.

philophilo:
@pbx  I can't seem to replicate the error. Was the `make setup` successful

pbx:
@philophilo there were no errors that I didn't report here. I'll try again later.

Makefile
```diff
@@ -81,14 +81,16 @@ target_help = \
...
+  'build - Builds the Docker image for django_accelerator' \
   'setup - Sets up a docker testing environent for django_accelerator' \
-  'pull-down - Stops the docker testing environent for django_accelerator' \
-  'run - runs [help, package, clean, code-check, coverage, coverage-run, ' \
-  'coverage-report, coverage-html-report, coverage-xml-report, ' \
+  'stop-server - Stops the Docker testing environent for django_accelerator' \
+  'run - runs Makefile commands inside the Docker container.' \
+  '\tSet the "command" variable to one of:' \
+  '\t [help, package, clean, code-check, coverage, coverage-run, ' \
+  '\t coverage-report, coverage-html-report, coverage-xml-report, ' \
```
pbx: This  help is very much improved, thanks. However there are still several targets here which don't run in the container. A couple examples:
```
$ make run command=coverage-html
... # Runs, produces summary table, then:
Couldn't get a file descriptor referring to the console
make: *** [coverage-html] Error 1
...
$ make run command=tox
...
ERROR: InterpreterNotFound: python3.7
```

Makefile
```diff
@@ -77,6 +77,20 @@ target_help = \
   'deploy - Deploy $$(release_name) to a $$(target).' \
   '\tValid targets include "staging" (the default), "production",' \
   '\t "test-1", and "test-2"' \
+  ' ' \
+  'install-sqlite - Installs sqlite3 on Travis from a script in' \
+  '\t scripts/run_make_commands_in_docker.sh, however it also can run in the' \
+  '\t accelerator container' \
+  'build - Builds the Docker image for django_accelerator' \
+  'setup - Sets up a Docker testing environent for django_accelerator' \
+  'stop-server - Stops the Docker testing environent for django_accelerator' \
```
pbx: This may be a confusing name, as the container is not a server. How about `stop-container`?

scripts/run_make_commands_in_docker.sh
```diff
@@ -0,0 +1,63 @@
...
+set_expected_commands (){
+    VARIABLES+=(
+        help
+        package
+        clean
+        code-check
+        coverage
+        coverage-run
+        coverage-report
+        coverage-html-report
```
pbx: Again, let's not include commands that don't work in the container.

scripts/run_make_commands_in_docker.sh
```diff
@@ -0,0 +1,63 @@
...
+for variable in ${VARIABLES[@]}; do
+    if [[ $1 == $variable ]]; then
+        echo "---running--"
+        make $1
+        success "Done..."
+        exit 0
+    fi
+done
+
+warning "[ make run command=<command> ] does not support \"$1\" command"
```
pbx: Thanks for adding this fallback message.

Makefile
```diff
@@ -77,6 +77,20 @@ target_help = \
   'deploy - Deploy $$(release_name) to a $$(target).' \
   '\tValid targets include "staging" (the default), "production",' \
   '\t "test-1", and "test-2"' \
+  ' ' \
+  'install-sqlite - Installs sqlite3 on Travis from a script in' \
```
pbx: The changes merged from [AC-7033](https://github.com/masschallenge/django-accelerator/pull/291) make it unnecessary to install an updated sqlite.  Right? Can't we remove the sqlite-related targets?

pbx:
@kaymation - as of [commit 9b2adca](https://github.com/masschallenge/django-accelerator/pull/293/commits/9b2adcafcaa3376668bdb0bb5cc842d1db6f1203) those have been renamed `bash-shell` and `stop-server` respectively. I'll update the PR text.

pbx:
**Status note:** in standup today, @philophilo agreed to post a detailed comment here responding to the unresolved issues raised in review above.  Once that's done we'll work together to figure out how to get it wrapped up.

philophilo:
@pbx I have removed both `tox` and `coverage-report` as they don't run successfully. `tox` fails with the error ` ERROR:  py36: InterpreterNotFound: python3.6`, both on the host machines and the within the Docker container. `coverage-html` fails with `Couldn't get a file descriptor referring to the console` since it requires to open a browser. After trying to use `xdg-open`, Docker still would need rights to open the browser on the host machine

pbx:
Thanks for the update @philophilo - what about the sqlite question?

philophilo:
@pbx I removed sqlite as well.

pbx:
@philophilo once you update the help I think we can get this merged. Given how rare it will be to want to manually `run` things inside this container, I think making the help text for that target more minimal would be fine.

PR #301, opened 2020-02-04
[AC-7310] Validate Judging Form when Judging Round is Made Active

jonkiparsky:
See sibling PR for testing notes


shakirandagire:
Doing QA for the task

PR #304, opened 2020-02-06
Merge master changes from v1.2.37 to development

pbx:
In producing the Django 2 release (v1.2.37), some problems with PRs led me to follow a slightly non-standard workflow, which resulted in changes being merged from `AC-6999` direct to master, which means we need to get those Django 2 changes into `development`.

Also see the impact-api PR. (There isn't one for accelerate, the AC-6999 changes have successfully been merged to development there.)


pbx:
Just waiting on an approval from @shanbady to merge this. (Hopefully the CC glitch will not block us. Still no word from their support on that)

PR #310, opened 2020-03-04
[AC-7397] Make UserRole.name unique in the db

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-7397

HOLD: this is breaking tests on accelerate, fixing that

NOTE: this is branched off AC-7514, so merging this will merge that. 

- verify that uniqueness has been applied as requested
- ask self whether we should be thinking about potential DB problems of arbitrarily making a field unique
- check accelerate and staging and see that the field is in fact unique there
- figure it's probably good enough
- run tests on accelerate and impact-api, just in case


jonkiparsky:
Closing due to test issues

PR #312, opened 2020-03-16
[AC-7397] Make UserRole name unique in DB

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-7397

Small change in the model, huge changes to tests. 
See sibling PR for changes to tests on accelerate. https://github.com/masschallenge/accelerate/pull/2564

NOTE: these PRs must merge/deploy simultaneously


iankigen:
Code looks good
QA looks good:
- Can only add user roles with unique names (/admin/mc/userrole/)

@jonkiparsky I think you should add a migration script to get rid of existing duplicate user roles to prevent the exception from ever occurring while creating a judging round(/staff/configure_judging_round/)

jonkiparsky:
@iankigen a migration to delete duplicate roles is a non-trivial piece of work if done seriously, since it would have to merge grants on duplicated roles. It is also not necessary since there are no existing duplicated roles on prod, staging or sandbox, which are the only machines whose data state we care about

accelerator/tests/contexts/context_utils.py
```diff
@@ -0,0 +1,11 @@
+from accelerator.tests.factories import UserRoleFactory
+from django.conf import settings
+import swapper
+
+app_label, model = swapper.split(settings.ACCELERATOR_USERROLE_MODEL)
+UserRole = swapper.load_model(app_label, model)
+
+
+def user_role_for_name(user_role_name):
```
geofrocker: I would prefer a more meaningful username such `get_or_create_user_role()`.  The current does not reveal why it exists or what it does.

geofrocker:
Code looks good. Waiting on travis

pbx:
Travis passed; I think the "Waiting for status" is an integration fail of some sort.

https://travis-ci.com/github/masschallenge/accelerate/builds/154846998

PR #314, opened 2020-04-02
[AC-7517] add ALUM to base startup role model

iankigen:
## [AC-7517](https://masschallenge.atlassian.net/browse/AC-7517)

**Changes Introduced:**
- add ALUM to base startup role model

**Testing**
- checkout to `AC-7517` in `django-accelerator` and `accelerate`.
- Navigate to any startup preview page with the following roles: [entrant, finalist, winner, or alumni]

Check for: 
- `Entrant` labels are visible to staff only
- All other labels are visible to staff, experts, finalists, and alumni 
- Under the heading Participation, display programs the startup in which the startup is/has been an entrant, finalist, winner, or alumni as list items under the heading
- Labels follow the format `year + status`, for example, **2017 Finalist**
- If `alumni` in `program_name`, finalist tag is not displayed in the same program
- The most recent is displayed first

[sibling PR](https://github.com/masschallenge/accelerate/pull/2572)

### DIFF
--- a/accelerator/tests/test_base_user_role.py
+++ b/accelerator/tests/test_base_user_role.py
@@ -3,7 +3,11 @@
 
 from django.test import TestCase
 
-from accelerator.tests.contexts import JudgeFeedbackContext
+from accelerator.tests.contexts import (
+    JudgeFeedbackContext,
+    UserRoleContext,
+)
+from accelerator_abstract.models.base_user_role import has_user_roles
 from accelerator.models import (
     has_user_role_base,
     UserRole,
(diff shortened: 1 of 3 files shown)


iankigen:
Correct. This can merge any time

geofrocker:
This should merge simultaneously with the sibling PR.

accelerator/tests/test_base_user_role.py
```diff
@@ -18,3 +22,9 @@ def test_has_user_role_base(self):
                                            UserRole.JUDGE,
                                            inactive_programs=True,
                                            active_or_ended_programs=True))
+
+    def test_has_user_roles(self):
```
jonkiparsky, line 26: Let's cover the negative case as well. (when user has user role "Mentor", `has_user_roles(user, [UserRole.ALUM, UserRole.FINALIST])' should return False)

PR #320, opened 2020-04-17
AC-7594: Add user role: Deferred Mentor

shakirandagire:
**Changes made to [AC-7594](https://masschallenge.atlassian.net/browse/AC-7594):**
- Added Deferred Mentor user role

**How to test:**
- Check out to this branch and sibling PR.
- Run `make migrate` to apply the migration
- Check [user role admin page](http://localhost:8181/admin/mc/userrole/) and notice that deferred mentor is one of the user roles.
- Visit [program role admin page](http://localhost:8181/admin/mc/programrole/?q=deferred+mentor) and view the past programs backfilled with the new user role

Sibling PR found on [accelerate](https://github.com/masschallenge/accelerate/pull/2592)


jonkiparsky:
@shakirandagire I see that this migration is numbered 0017 and depends on 0016, but there are already migrations 0017 and 0018 in the stack. The easiest way to fix this is to renumber this migration by changing the file name to start with 0019 and cause it to depend on 0018 (edit the entry in `dependencies`) 
Let me know if you have questions about this

shakirandagire:
@jonkiparsky Let me do that and update accordingly

shakirandagire:
@jonkiparsky I have made the changes to the naming of the migration

accelerator/migrations/0019_add_deferred_user_role.py
```diff
@@ -0,0 +1,32 @@
...
+    Program = apps.get_model('accelerator', 'Program')
+    ProgramRole = apps.get_model('accelerator', 'ProgramRole')
+    user_role = UserRole.objects.create(
+        name=DEFERRED_MENTOR, sort_order='17')
+    for program in Program.objects.all():
+        try:
+            ProgramRole.objects.get_or_create(
+                program=program,
+                user_role=user_role)
+        except IntegrityError:
```
jonkiparsky, line 19: I'm curious about the IntegrityError. Did you have any idea of what was causing this to be raised?

shakirandagire: When I ran `make migrate` I was getting an `django.db.utils.IntegrityError: (1062, "Duplicate entry '' for key 'name'")` and I realized that when a program role is created but the name is empty so that is why it is triggered.

pbx:
After we deploy this, we'll want to check that the migration 0019 ran successfully. After the deploy of v1.2.56 we noticed that `bin/django showmigrations` reported 0017 and 0018 as _not_ having run, though inspection of the database indicates that they did.

Hopefully the release containing this ticket will end in a good state with all migrations applied (and recorded as such in the migrations table). If not, we'll make a P2 bug ticket to investigate and fix.

(I'll leave this note on the ticket as well.)

shakirandagire:
@pbx thanks for the heads up on the migration issue.

accelerator/migrations/0019_add_deferred_user_role.py
jonkiparsky: Ah, then we need to deal with that. 

On clean-db, before migration:
```
In [3]: mc.ProgramRole.objects.count()
Out[3]: 1340
```
after migration: 
```
In [4]: mc.ProgramRole.objects.count()
Out[4]: 1341
```


Only one was created, and it was:
```
In [6]: mc.ProgramRole.objects.latest("id")
Out[6]: <ProgramRole: >
```
hence the integrity errors - it was trying to create multiple Program Roles with the same (blank) name. With the `except: pass`, we just don't create the Program Roles at all, which is not what we want. 

I suggest composing a name, such as
```
    name="{} {} Deferred Mentor".format(program.start_date.year, program.program_family.name)
```

(ideally, we'd be able to use the Program.url_slug field, but staff have not abided by our standards, so we cannot get the year from that slug as we'd expect to)

shakirandagire: This explains it much better. Thanks for the suggestion. Trying it out currently

iankigen:
@shakirandagire you have lint errors

accelerator/migrations/0019_add_deferred_user_role.py
```diff
@@ -12,16 +11,15 @@ def add_deferred_user_role(apps, schema_editor):
     user_role = UserRole.objects.create(
         name=DEFERRED_MENTOR, sort_order='17')
     for program in Program.objects.all():
-        name="{} {} Deferred Mentor".format(
-            program.start_date.year, 
+        name = "{} {} Deferred Mentor".format(
+            program.start_date.year,
             program.program_family.name)
-        ProgramRole.objects.get_or_create(
+        return ProgramRole.objects.get_or_create(
```
iankigen, line 17: I don't think you should have a return here

shakirandagire:
@iankigen 
I have fixed the linting issues.

accelerator/migrations/0019_add_deferred_user_role.py
shakirandagire: Let me remove it.

iankigen:
test coverage drop is on a migration file, is there a way to ignore migration files test coverage?

PR #326, opened 2020-06-05
AC-7818 Add 'Meeting Information' to Office Hours model

geofrocker:
### Changes introduced in: [AC-7818](https://masschallenge.atlassian.net/browse/AC-7818):
- Add a migration to add meeting_info
### How to test
- Checkout this branch
- On accelerate `make migrate`
- Check out this branch too on impact and run `make runserver`
- Schedule office hours as a mentor and and navigate to this [api endpoint](http://localhost:8000/api/v1/office_hours_calendar/)
- Notice the new field `meeting_info`
### This PR should merge before the one on impact


accelerator_abstract/models/base_mentor_program_office_hour.py
```diff
@@ -49,6 +49,7 @@ class BaseMentorProgramOfficeHour(AcceleratorModel):
         on_delete=models.CASCADE)
     notify_reservation = models.BooleanField(default=True)
     topics = models.CharField(max_length=500, blank=True)
+    meeting_info = models.CharField(max_length=100, blank=True, default="")
```
iankigen, line 52: according to the definition of done:
Limit of 256 characters with validation message: "Meeting information cannot exceed 256 characters."

geofrocker: Added the limit. Django handles validation with a clear validation message. I see no need to override that. @iankigen

iankigen, line 52: `max_length` was only changed on the migration file but not the model

freshprincekla: Fixed this as @geofrocker isn't in today

PR #331, opened 2020-06-24
AC-7723 Too-long 'topics' field in OfficeHoursReservationView should not throw a 500

geofrocker:
### Changes Introduced in [AC-7723 ](https://masschallenge.atlassian.net/browse/AC-7723 ):
- Remove the character limit on the topics field
### Testing:
- Check out this branch
- Visit the office hours page for Boston 2019 [here](http://localhost:8181/officehours/bos/2019/)
- Schedule office hours
- Try reserving office hours for a finalist.
- Notice that there is a limit of 2000 chars on Meeting information

### DIFF
--- a/accelerator_abstract/models/base_mentor_program_office_hour.py
+++ b/accelerator_abstract/models/base_mentor_program_office_hour.py
@@ -48,7 +48,7 @@ class BaseMentorProgramOfficeHour(AcceleratorModel):
         blank=True,
         on_delete=models.CASCADE)
     notify_reservation = models.BooleanField(default=True)
-    topics = models.TextField(blank=True)
+    topics = models.CharField(max_length=2000, blank=True, default="")
     meeting_info = models.CharField(max_length=256, blank=True, default="")
 
     class Meta(AcceleratorModel.Meta):
--- a/accelerator_abstract/models/base_mentor_program_office_hour.py
+++ b/accelerator_abstract/models/base_mentor_program_office_hour.py
@@ -48,7 +48,7 @@ class BaseMentorProgramOfficeHour(AcceleratorModel):
         blank=True,
         on_delete=models.CASCADE)
     notify_reservation = models.BooleanField(default=True)
-    topics = models.CharField(max_length=500, blank=True)
+    topics = models.TextField(blank=True)
     meeting_info = models.CharField(max_length=256, blank=True, default="")
 
     class Meta(AcceleratorModel.Meta):
--- a/accelerator/migrations/023_alter_topics_field_office_hours.py
+++ b/accelerator/migrations/023_alter_topics_field_office_hours.py
@@ -13,6 +13,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='mentorprogramofficehour',
             name='topics',
-            field=models.TextField(blank=True),
+            field=models.CharField(blank=True, default='', max_length=2000),
         ),
     ]
--- a/accelerator/migrations/023_alter_topics_field_office_hours.py
+++ b/accelerator/migrations/023_alter_topics_field_office_hours.py
@@ -1,0 +1,18 @@
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accelerator', '0022_add_meeting_info_on_office_hour_model'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='mentorprogramofficehour',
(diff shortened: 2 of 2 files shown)


jonkiparsky:
Drive-by comment: per Paul's comment in standup, if this data is passed through the URL then this is not the fix we want. (since urls can be length-limited) 

Check with @hfreund but I would suggest adding notifying the user of the length restriction and checking the field before submit.

geofrocker:
> Drive-by comment: per Paul's comment in standup, if this data is passed through the URL then this is not the fix we want. (since urls can be length-limited)
> 
> Check with @hfreund but I would suggest adding notifying the user of the length restriction and checking the field before submit.
### I have raised the new  PR on accelerate
It seems users would like to pass more information. Do we know the maximum length that can be passed in the url or we keep the 500 restriction? cc @pbx @jonkiparsky

pbx:
@geof you make a good observation that we have good evidence users are wanting more than 500 characters! I'll leave it to Product to make the call what to do there.

As for URL length, looks like 2000 characters is a commonly used limit, based on MSIE's constraint. https://support.microsoft.com/en-us/help/208427/maximum-url-length-is-2-083-characters-in-internet-explorer

geofrocker:
Great. @pbx Wouldn't it make sense to use a text field(remove restriction) on the backend and do the validation on the frontend?

pbx:
@geofrocker I could see that path working, yeah. As long as the back-end code isn't depending on the front-end code to prevent errors (i.e. I wouldn't consider adding a front-end limit of 500 characters to the current system, to be sufficient). The Product/UX question of how large that field should be -- that's for @hfreund to weigh in on.

hfreund:
@geofrocker is this character limit on the topics field going to apply to the new office hours interface as well or will it only work for the current, soon-to-be-dead interface?

geofrocker:
@hfreund It would have to cut across. Both interfaces are potentially affected by this bug cc @jonkiparsky

hfreund:
@geofrocker So if we are making this change for both my feeling is that it makes more sense to do this on the backend since any frontend ui notification on the existing office hours will not be reusable on the new office hours, however, a backend validation can trigger the appropriate front-end notification for both interfaces.

Regarding length, I think we should set a high limit here unless there are good reasons not to do so. Some of our mentors like to write a lot - we shouldn't throw errors at them if we don't need to. A standard Zoom invite from MC is just under 1,000 characters. If we want to allow that and some space for a person to write a message, I would say 1,500 is the minimum ceiling we could reasonably specify. Higher would be better if there is no performance hit.

geofrocker:
I guess the way forward here is to:
- Add backend validation with the limit set to 2000
- Implement frontend validation for both the new and old office hours to the same limit.
cc @pbx @jonkiparsky

accelerator_abstract/models/base_mentor_program_office_hour.py
```diff
@@ -48,7 +48,7 @@ class BaseMentorProgramOfficeHour(AcceleratorModel):
         blank=True,
         on_delete=models.CASCADE)
     notify_reservation = models.BooleanField(default=True)
-    topics = models.CharField(max_length=500, blank=True)
+    topics = models.TextField(blank=True)
```
iankigen, line 51: `topics` field type is inconsistent with the migration file otherwise LGTM.
```
field=models.CharField(blank=True, default='', max_length=2000),

jonkiparsky:
Not to complicate things further, but I believe 2000 characters is the hard limit for the URL, so a limit of 2000 characters probably not helpful. 

From a [discussion on quora](https://www.quora.com/What-is-the-max-length-for-URLs-in-each-browser): 
> Microsoft Internet Explorer (Browser)
>  Microsoft states that the maximum length of a URL in Internet Explorer is 2,083 characters, with no more than 2,048 characters in the path portion of the URL.

hfreund:
> From a [discussion on quora](https://www.quora.com/What-is-the-max-length-for-URLs-in-each-browser):
> 
> Microsoft Internet Explorer (Browser)
> Microsoft states that the maximum length of a URL in Internet Explorer is 2,083 characters, with no more than 2,048 characters in the path portion of the URL.

FWIW we don't support IE.

accelerator_abstract/models/base_mentor_program_office_hour.py
geofrocker: aaah. I see thanks

pbx:
@hfreund to clarify - I didn't bring up IE because it's one of our supported browsers, but because its URL length limitation has become a de facto standard for maximum URL length.

PR #335, opened 2020-07-09
[AC-7750-fix]: Fix migrations causing test failures

shakirandagire:
Changes introduced in AC-7750-fix:
Fixed the migration issue.

How to test:
- Check that there are no migration with same number
- Check that the tests are passing as required.


shakirandagire:
@pbx and @geofrocker the tests are now passing in travis

PR #345, opened 2020-08-14
[AC-8075] Update imports of DjangoModelFactory

jonkiparsky:
This PR updates imports of DjangoModelFactory, which was changed in a recent FactoryBoy release

Release notes: https://factoryboy.readthedocs.io/en/latest/changelog.html#id2

If tests pass, this should be good.


jonkiparsky:
@iankigen lint has been linted

PR #347, opened 2020-08-31
[AC-8081] Add migration to add active user role fields

geofrocker:
Follow test instructions on the accelerate PR


accelerator_abstract/models/base_nav_tree_item.py
```diff
@@ -33,6 +33,7 @@ class BaseNavTreeItem(TreeItemBase, AcceleratorModel):
         blank=True)
     active_program = models.BooleanField(default=False)
     display_single_item = models.BooleanField(default=True)
+    active_user_roles = models.BooleanField(default=False)
```
iankigen, line 36: why not use the existing `active_program` boolean field?

PR #348, opened 2020-09-29
AC-8169 :: Create Minimal Expert Navigation Tree and Landing Page

freshprincekla:
**Changes introduced in [AC-8169](https://masschallenge.atlassian.net/browse/AC-8169)**
- Add user_type to the nav tree items to enable expert only side and sub-nav items
- Add a redirect to expert overview landing page. `/dashboard/expert/overview/`

**How to test**
Checkout this PR and the sibling [PR](https://github.com/masschallenge/django-accelerator/pull/348)
**Test the redirect**
- Create a new fluent page [`Expert`](http://localhost:8181/admin/fluent_pages/page/add/) and publish it as a `category header` as shown below:
Screen Shot 2020-09-29 at 3 27 15 PM
Create another fluent page [`overview`](http://localhost:8181/admin/fluent_pages/page/add/) and publish it as a `page` as shown below:
Screen Shot 2020-09-29 at 3 33 36 PM
- After the Category header and page have been created, navigate to the [fluent page admin](http://localhost:8181/admin/fluent_pages/) drag and drop the  overview page on the the just created Category header.

- Masquerade as an expert and see that you are redirected to `/dashboard/expert/overview/`

**Test that side nav items can be displayed according to user_type**

To ensure that the side nav items can be displayed according to user_type
- As demoadmin, go to this side nav item http://localhost:8181/admin/mc/navtree/1/change/item_82/ and set the user_type to Entrepreneur
- Masquerade as this  [expert](http://localhost:8181/admin/simpleuser/user/56778/masquerade/) user 
- See that on the side nav the item isn't displayed.


accelerator_abstract/models/base_core_profile.py
```diff
@@ -207,7 +211,17 @@ def _get_staff_landing_page(self):
         if has_staff_clearance(self.user):
             return '/staff'
 
+    def _has_expert_nav_enabled(self):
+        bt = BulletTrain(environment_id="aX45EUqSsAqhTvv5nW7WEL")
```
jonkiparsky, line 215: I don't love this magic number. I would prefer that it were a constant with a name telling me what it means.
Or better still, I'd like to have a function which consults bullet train and asks whether some `feature_name` is enabled. That way when we delete this code as part of the cleanup ticket, the next person to use feature flagging doesn't have to reinvent this wheel.

accelerator_abstract/models/base_nav_tree_item.py
```diff
@@ -5,6 +5,11 @@
 from sitetree.models import TreeItemBase
 
 from accelerator_abstract.models.accelerator_model import AcceleratorModel
+EXPERT_USER_TYPE = 'EXPERT'
+ENTREPRENEUR_USER_TYPE = 'ENTREPRENEUR'
+
+USER_TYPES = ((EXPERT_USER_TYPE, 'Expert'),
```
jonkiparsky, line 11: EXPERT_USER_TYPE and ENTREPRENEUR_USER_TYPE are already defined in base_base_profile.py and can be imported. USER_TYPES is also defined there, but since it includes the MEMBER_USER_TYPE which we don't want to expose here, we should probably redefine it here. In order to prevent future confusion, the name should be distinct.

jonkiparsky (changes requested):

shakirandagire (approved):
QA looks good to me. 
Details for what was done is on https://github.com/masschallenge/accelerate/pull/2768

accelerator/utils.py
```diff
@@ -14,3 +17,11 @@ def create_mc_permission(permission):
         group.permissions.add(new_perm)
     for user in permission.user_set.all():
         user.user_permissions.add(new_perm)
+
+
+def bullet_train_access_util(feature_name):
```
jonkiparsky, line 22: I don't love this name. "Access" to me suggests something like a permission check, and "util" suggests an object, not a function (I know we've already made the latter mistake once in the codebase, but let's not make it a habit!)

Something like `bullet_train_feature_check` would be preferable for me.

jonkiparsky (approved):

PR #353, opened 2020-10-20
adding fixes for merged impact tests

shanbady:
See associated PR: https://github.com/masschallenge/accelerate/pull/2789


pbx (changes requested):
I agree with Travis:

`accelerator/tests/utils.py:22:5: E722 do not use bare except'`

pbx (approved):

PR #355, opened 2020-10-27
[Ac 8296] Remove character limit on application answer_text

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-8296


freshprincekla (approved):

jonkiparsky:
QA: I was able to set a text limit to 5K characters, and save an answer of that length.

iankigen (approved):

PR #360, opened 2020-11-30
[AC-8363] Add birth_year field to core profile

iankigen:
## [AC-8363](https://masschallenge.atlassian.net/browse/AC-8363)

**Changes introduced**
- add birth_year field to core profile

**Testing**
- checkout to this branch
- After [AC-7802](https://github.com/masschallenge/accelerate/pull/2800), running `make migrate` shows migrations have run successfully, however, it does not update the db tables. As an alternative, you can apply migrations using the impact API container. Running migration using accelerate container does not work.
   - on `impact-api/docker-compose.yml`. comment out `assets`, `front-end`, `start_dependencies` definition. Also remove `assets` and `front-end` from `web:... depends_on:` (optional - I've mentioned this just in case you need to save some build time)
   - To apply migrations: `docker-compose run --rm web ./manage.py migrate accelerator`
- Confirm that:
  - `birth_year` exists on `EntrepreneurProfile`, `ExpertProfile` and `MemberProfile` models
  - core_profile_factory to includes `birth_year` (default to null)
  - `birth_year` is a DateField
  - `birth_year` not required


iankigen:
Do not merge before the migration issue is fixed

accelerator/migrations/0027_add_birth_year_to_core_profile.py
```diff
@@ -0,0 +1,28 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accelerator', '0026_remove_character_limit_to_application_answer'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='entrepreneurprofile',
```
geofrocker, line 14: this should be on the core_profile model

geofrocker (changes requested):

accelerator/migrations/0027_add_birth_year_to_core_profile.py
iankigen: it is specified on the `base_core_profile` model, however, since `base_core_profile` is abstract and has no table on the db, the migrations are created for the child classes which in this case is `entrepreneurprofile`, `expertprofile` and `memberprofile`.

freshprincekla (approved):
Code changes look good,

freshprincekla:
> Do not merge before the migration issue is fixed

is this still valid or was the issue fixed?

geofrocker (approved):
QA looks good
- I am able to see the applied migration

PR #361, opened 2020-12-01
[AC-8366] Add Ethno-racial Identity Field to core_profile model

iankigen:
## [AC-8366](https://masschallenge.atlassian.net/browse/AC-8366)

**Changes introduced**
- Add Ethno-racial Identity Field to core_profile model

**Testing**
- After [AC-7802](https://github.com/masschallenge/accelerate/pull/2800), running `make migrate` shows migrations have run successfully, however, it does not update the db tables. As an alternative, you can apply migrations using the impact API container. Running migration using accelerate container does not work.
   - on `impact-api/docker-compose.yml`. comment out `assets`, `front-end`, `start_dependencies` definition. Also remove `assets` and `front-end` from `web:... depends_on:` (optional - I've mentioned this just in case you need to save some build time)
   - To apply migrations: `docker-compose run --rm web ./manage.py migrate accelerator`
- Confirm that:
  - a new model `ethno_racial_identity` is added to accelerate 
  - a new field `ethno_racial_identification` is added to the core_profile model as a many-to-many link to ethno_racial_identity
  - a new Boolean field `authorization_to_share_ethno_racial_identity` is added to the `core_profile` model, defaulting to `False`.
  - a schema migration exists to make these changes to the DB structure
  - the initial list of identity options is added to the database via a data migration
  - the appropriate factories are updated with a sensible default


accelerator_abstract/models/base_ethno_racial_identity.py
```diff
@@ -0,0 +1,45 @@
+from django.db import models
+
+from accelerator_abstract.models.accelerator_model import AcceleratorModel
+
+BLACK = 'Black'
```
jonkiparsky, line 5: Why are these specified in the model? The intent was to keep this information in the database, and we should never need to consult these values within the app.

accelerator_abstract/models/base_ethno_racial_identity.py
```diff
@@ -0,0 +1,45 @@
...
+    (WHITE_CAUCASIAN, 'White or Caucasian'),
+    (PREFER_NOT_TO_STATE, 'I prefer not to say'),
+]
+
+
+class BaseEthnoRacialIdentity(AcceleratorModel):
+    identity = models.CharField(max_length=100,
+                                blank=True,
+                                null=True,
+                                choices=ETHNO_RACIAL_IDENTITY_CHOICES)
```
jonkiparsky, line 38: Again why are we restricting choices to a hard-coded set?

jonkiparsky (changes requested):

freshprincekla (approved):
- ethno_racial_identity model was added to accelerate ✅ 
- ethno_racial_identification was added to the core_profile model ✅ 
- authorization_to_share_ethno_racial_identity was added to the core_profile model, defaulting to False. ✅ 
- the appropriate factories were updated with a sensible default ✅

jonkiparsky (approved):

PR #364, opened 2020-12-09
AC-8380 - reset migrations

pbx:
### Review suggestions
1. Check out the branch
2. Load the new clean database: `make load-db db_name=initial_schema_AC8380`
3. Confirm that the migrations ran and that your local environment is in a usable state (e.g. `make run-server` and log in)

Also look at the remaining migrations and confirm that their dependencies seem correct

This PR plus the new clean database are the meat of this ticket's deliverables, but **also see [sibling accelerate PR](https://github.com/masschallenge/accelerate/pull/2827)** for the updated `clean.sql` (needed in order to generate clean DB dumps).

The "WIP" and "Don't merge" labels are attached because we don't yet have a deployment strategy for prod.


shanbady (changes requested):
I dont see a new initial migration file (0001_initial) for django-accelerator. does this need to be generated?

pbx:
@shanbady - it's there - https://github.com/masschallenge/django-accelerator/pull/364/files#diff-0793a6ce81e43684655ac99d53f1de14da6d1513e7285ba4ae519cac885db3d6

shanbady:
Ah ok - i confused it with being a file removal. looks good to me in that case.

shanbady (approved):

pbx:
Closing this PR and moving the ticket back to To Do, because the problem in [AC-8411](https://masschallenge.atlassian.net/browse/AC-8411) (migrations aren't having effect when run on AWS) blocks it.

PR #366, opened 2020-12-17
[AC-8419] P0: Django 2.2  DateTime fields are incompatible with MySQL 5.5

iankigen:
## [AC-8419](https://masschallenge.atlassian.net/browse/AC-8419)
**Changes introduced**
- Update MySQL data type mapper to use `datetime` instead of `datetime(6)`
**Testing**
- ssh into test3
- Confirm that you can run migrations without MySQL compatibility error: `django.db.utils.ProgrammingError: (1064, "You have an error in your SQL syntax;`
[Sibling PR](https://github.com/masschallenge/accelerate/pull/2833)

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -11,3 +11,4 @@ tox
 mock
 factory-boy
 swapper
+mysqlclient
--- a/settings.py
+++ b/settings.py
@@ -4,6 +4,9 @@
 import os
 import sys
 import logging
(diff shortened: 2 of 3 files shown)


jonkiparsky (approved):
I believe this should be okay to merge - the code is straightforward and we've tested it on an AWS stack.

pbx (approved):
Approved, as noted on the sibling PR

PR #379, opened 2021-01-19
AC-8365 Add demographic fields to Entrepreneur profile

geofrocker:
Follow the testing instruction on the accelerate PR


iankigen:
@pbx it seems we forgot to restore: https://masschallenge.atlassian.net/browse/AC-8400 which had set blank=True making this PR necessary

pbx (approved):
Looks correct. We use this technique in many other models.

iankigen (approved):
Looks good

PR #381, opened 2021-01-22
[AC-8490] Migrate `gender` data to `gender_identity`

iankigen:
## [AC-8490](https://masschallenge.atlassian.net/browse/AC-8490)

**Changes introduced**
- Add data migrations to migrate `gender` data to `gender_identity`

**Testing**
- Checkout to this branch and `$ make migrate`
- Confirm that all existing gender data is migrated to the gender_identity object according to the acceptance criteria on [AC-8490](https://masschallenge.atlassian.net/browse/AC-8490)


accelerator/migrations/0033_migrate_gender_data.py
```diff
@@ -0,0 +1,66 @@
+# Generated by Django 2.2.10 on 2021-01-22 12:13
+import sys
```
jonkiparsky: why is this imported?

jonkiparsky (changes requested):
One comment about performance (which may have gotten lost, thanks github) 

Higher-level comment about the desirability of this migration:
When we originally discussed the rollout of this, we decided that we would not be migrating the original data, since we would rather have users revisit this and update their identity. 
Why have we changed this decision?
@hfreund  please clarify

accelerator/migrations/0033_migrate_gender_data.py
iankigen: removed. I was using `sys.stdout` to display progress % on my dev environment

accelerator/migrations/0033_migrate_gender_data.py
```diff
@@ -0,0 +1,64 @@
...
+def add_gender_identity(profile, gender_choice):
+    if not profile.gender_identity.filter(pk=gender_choice.pk).exists():
+        profile.gender_identity.add(gender_choice.pk)
+
+
+def migrate_gender_data_to_gender_identity(apps, schema_editor):
+    users = get_user_model().objects.all()
+    gender_choices = get_gender_choice_obj_dict(apps)
+
+    for user in users:
```
jonkiparsky, line 42: Why are we looping over users and then calling get_profile? Why not just loop over all of the profiles?
For that matter, if it's possible to do a bulk update of gender_identity, why not filter profiles by their original gender and bulk-update them?

hfreund:
> When we originally discussed the rollout of this, we decided that we would not be migrating the original data, since we would rather have users revisit this and update their identity.
> Why have we changed this decision?

@jonkiparsky several factors caused a change in direction:
- There is a strong desire by the business to move to the new `gender_identity` field as quickly as possible; both the business and data team prefer minimizing the amount of time that two types of similar data sets exist.
- Overall delay of demographic data epic put us 4+ weeks behind the original target. A lot of new accounts were created in that time using the old gender field and the business would like them to be on the new field.
- The original plan to prompt users to take some action to review their profile and update their demographic information is partially scoped but not yet planned. The delays in shipping the demographic work cut into our window to build the follow-on deferrable modal. As other business priorities crop up there may not be enough bandwidth to deliver the deferrable modal soon enough that a majority of active users update their data to the point where we can switch all processes (i.e. judging & reporting) to using `gender_identity` instead of `gender`.

jonkiparsky (approved):
Confirmed that migration works as expected.

PrinceDavis (approved):
Migration ran successfully ✅

PR #392, opened 2021-03-03
[AC-8416] Add UserDeferrableModal model

iankigen:
## [AC-8416](https://masschallenge.atlassian.net/browse/AC-8416)

**Changes introduced**
- Add UserDeferrableModal model and factory

**Testing**
- Confirm this PR meets the definition of done according to [AC-8416](https://masschallenge.atlassian.net/browse/AC-8416)


geofrocker:
failing code climate

accelerator_abstract/models/base_user_deferrable_modal.py
```diff
@@ -0,0 +1,25 @@
...
+            AcceleratorModel.Meta.app_label, 'DeferrableModal'),
+        on_delete=models.CASCADE)
+    is_deferred = models.BooleanField(default=False)
+    deferred_to = models.DateTimeField(null=True, blank=True)
+
+    class Meta(AcceleratorModel.Meta):
+        abstract = True
+        verbose_name = 'User Deferrable Modal'
+
+    def __str__(self):
```
geofrocker, line 24: Since we are not adding this to the admin interface, I guess we don't need this now.

geofrocker (changes requested):

accelerator_abstract/models/base_user_deferrable_modal.py
iankigen: This was mostly meant to provide a more readable output when working with the object on the shell and not necessarily for the admin interface

geofrocker: Non-blocking then. Instead of self.pk, how about we return self.defferable_modal.name?

kenware (approved):
Changes look goods

accelerator_abstract/models/base_user_deferrable_modal.py
jonkiparsky: Chiming in on this, you would typically use `__repr__` rather than `__str__` for the developer-facing representation of the object. `__str__` would be used, for example, if we wanted a string that could be usefully evaluated in a template context.

geofrocker (approved):
QA Looks good
- The modal has been added and works as expected

jdaw112105:
pineapple soup.

PR #393, opened 2021-03-04
AC-8644 :: P1: Expert Overview displays to finalists

freshprincekla:
**Changes introduce in [AC-8644](https://masschallenge.atlassian.net/browse/AC-8644)**
- added a user_type flag to the UserRoleMenu
- filtered associations when user_type is specified

**How to test**
- On test4, as an admin, go to the expert user role menu. 
- You will see a selection box for user_type. 
- Select Expert and save
- You can then masquerade as an entrepreneur and see that you no longer see the `Expert Overview`
- Masquerade as an expert and see that you see the `Expert Overview`


jonkiparsky (approved):
This all looks fine. My only question is whether there are any existing UserRoleMenu objects that need to have user_type set (if so, we'd want a data migration for that)

kenware (approved):

kenware:
Code looks good

PR #407, opened 2021-05-04
[AC-8802] Ensure that CoreProfile has functions needed for Startup membership/ownership

iankigen:
## [AC-8802](https://masschallenge.atlassian.net/browse/AC-8802)

**Changes introduced**
- move functions needed for Startup membership/ownership to CoreProfile

**Testing**
- Confirm that the following def of done conditions are met:
  - move is_alum_in_residence and first_startup onto CoreProfile
  - make add no-op get_active_alerts to EntrepreneurProfile
  - Result of this ticket: EntrepreneurProfile defines no functions

[Sibling PR](https://github.com/masschallenge/accelerate/pull/2959)

### DIFF
--- a/accelerator/tests/test_core_profile.py
+++ b/accelerator/tests/test_core_profile.py
@@ -270,9 +270,8 @@ class TestCoreProfile(TestCase):
 
 def _user_is_alum_in_residence(air_program=None, non_air_program=None):
     user = EntrepreneurFactory()
-    if air_program:
-        user = EntrepreneurFactory(profile__current_program=air_program)
-    context = UserRoleContext(BaseUserRole.AIR, user=user)
+    context = UserRoleContext(
+        BaseUserRole.AIR, user=user, program=air_program)
     user_profile = context.user.get_profile()
     program_of_interest = non_air_program or air_program
     if program_of_interest:
--- a/accelerator/tests/test_core_profile.py
+++ b/accelerator/tests/test_core_profile.py
@@ -124,7 +124,7 @@ class TestCoreProfile(TestCase):
         self.assertTrue(
             profile.interest_category_names() == [interest_category.name])
 
-    def test_is_alum_in_residence(self):
+    def test_expert_is_alum_in_residence(self):
         user = expert(BaseUserRole.AIR)
         profile = user.get_profile()
         self.assertTrue(profile.is_alum_in_residence())
--- a/accelerator/tests/test_core_profile.py
+++ b/accelerator/tests/test_core_profile.py
@@ -253,3 +253,28 @@ class TestCoreProfile(TestCase):
         user_profile = EntrepreneurProfileFactory(landing_page="/")
         landing_page = user_profile.check_landing_page()
         self.assertEqual(landing_page, user_profile.default_page)
+
+    def test_is_alum_in_residence(self):
+        self.assertTrue(_user_is_alum_in_residence())
+
+    def test_is_alum_in_residence_returns_true_if_in_program(self):
+        air_program = ProgramFactory()
+        self.assertTrue(_user_is_alum_in_residence(air_program))
+
+    def test_is_alum_in_residence_returns_false_if_not_in_program(self):
+        air_program = ProgramFactory()
+        non_air_program = ProgramFactory()
+        self.assertFalse(_user_is_alum_in_residence(
+            air_program, non_air_program))
+
+
+def _user_is_alum_in_residence(air_program=None, non_air_program=None):
+    user = EntrepreneurFactory()
+    if air_program:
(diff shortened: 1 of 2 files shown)


iankigen:
moving tests from accelerate to django-accelerator to fix the coverage change

accelerator_abstract/models/base_core_profile.py
```diff
@@ -187,10 +188,18 @@ def is_program_graduate(self, program=None):
...
 
-    def is_alum_in_residence(self):
-        return self.user.programrolegrant_set.filter(
-            program_role__user_role__name=BaseUserRole.AIR
-        ).exists()
+    def is_alum_in_residence(self, program=None):
+        if self.user_type == ENTREPRENEUR_USER_TYPE.lower():
+            qs = self.user.programrolegrant_set.filter(
+                program_role__user_role__name=BaseUserRole.AIR,
+                program_role__program__exact=self.current_program)
```
jonkiparsky, line 195: This is interesting. What this specifies is that we're only interested in AIR grants for the user's `current_program`. But that field is not maintained in any way that I'm aware of, so this means this is probably incorrect. 

If we were to remove the filter on `current_program`, that would mean that an entrepreneur user is an AIR if they have an AIR grant, period. (this is already true for non-entrepreneurs) I'm not sure why this would be a bad thing. 
@hfreund Can you think of any product reason why we'd care about current_program here?

accelerator_abstract/models/base_core_profile.py
```diff
@@ -187,10 +188,18 @@ def is_program_graduate(self, program=None):
             qs = qs.filter(program_role__program=program)
         return qs.exists()
 
-    def is_alum_in_residence(self):
-        return self.user.programrolegrant_set.filter(
-            program_role__user_role__name=BaseUserRole.AIR
-        ).exists()
+    def is_alum_in_residence(self, program=None):
+        if self.user_type == ENTREPRENEUR_USER_TYPE.lower():
```
jonkiparsky, line 192: Sooner or later we're going to eliminate this distinction. If we stop caring about `current_program` (see below) then the only difference this makes is that for one set of users we can specify a program and get back an answer to the question "is this user an AIR in this program?".
I don't know if we're actually asking this question in the current codebase, but it seems like a reasonable thing to allow here and I don't see why we shouldn't allow it for all user types.

accelerator/tests/test_core_profile.py
```diff
@@ -253,3 +253,28 @@ def test_landing_page_if_profile_landing_page_is_set_to_home(self):
...
+        air_program = ProgramFactory()
+        non_air_program = ProgramFactory()
+        self.assertFalse(_user_is_alum_in_residence(
+            air_program, non_air_program))
+
+
+def _user_is_alum_in_residence(air_program=None, non_air_program=None):
+    user = EntrepreneurFactory()
+    if air_program:
+        user = EntrepreneurFactory(profile__current_program=air_program)
```
jonkiparsky, line 274: Any time a new test requires setting `current_program`, that should alarm us. Let's see if we can't get rid of this. (see comment on `base_core_profile.py`)

jonkiparsky (changes requested):
Overall, looks good with a few minor issues.

accelerator_abstract/models/base_core_profile.py
```diff
@@ -228,13 +230,9 @@ def is_partner_admin(self):
             partner_administrator=True).exists()
 
     def get_active_alerts(self, page=None):
-        """Return any active alerts for the user, that are relevant for
-        the current 'page' of the application.
-        May be overridden by subclasses (e.g., ExpertProfile,
-        EntrepreneurProfile, etc.)
+        """no op
```
kenware, line 233: What does this comment mean? we may remove all the comment if it does not mean anything reasonable

iankigen: it means no operation.

kenware: Since we do not need to override this function in the EntrepreneurProfile, do we still need it here?

kenware (approved):
Looks good

- [X] move is_alum_in_residence and first_startup onto CoreProfile
- [X] make add no-op get_active_alerts to EntrepreneurProfile
- [X] Result of this ticket: EntrepreneurProfile defines no functions

accelerator_abstract/models/base_core_profile.py
jonkiparsky: @kenware The purpose of this effort is to ensure that EntrepreneurProfile does not define any functionality or fields - it should inherit everything from CoreProfile.

jonkiparsky (approved):

PR #410, opened 2021-05-11
[AC-8824]: Update location_regional field with custom migration

kenware:
[AC-8824](https://masschallenge.atlassian.net/browse/AC-8824)

How to test:
- Confirm that the migration runs on your local machine without error
- If necessary, you can run an SQL on accelerator_startup on a field `location_regional`; to ensure that Items changed are persisted
-


accelerator/migrations/0041_correct_data_in_accelerator_startup.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 2.2.16 on 2021-05-11 15:00
+
+from django.db import migrations
+
+
+MASSACHUSETTS = 'Massachusetts'
+CALIFONIA = 'California'
```
jonkiparsky, line 7: Typo, s/b `CALIFORNIA`

accelerator/migrations/0041_correct_data_in_accelerator_startup.py
```diff
@@ -0,0 +1,36 @@
...
+NEW_YORK = 'New York'
+RHODE_ISLAND = 'Rhode Island'
+CALIFONIA_DATA_LIST = [
```
jonkiparsky, line 10: Shouldn't this be `MASSACHUSETTS_DATA_LIST`?

jonkiparsky (changes requested):
Two trivial issues

accelerator/migrations/0041_correct_data_in_accelerator_startup.py
kenware: Thanks for spotting these @jonkiparsky

jonkiparsky (approved):
Reviewing the data on prod and the [dataset used by the CRS library](https://github.com/country-regions/country-region-data/blob/master/data.json) I see that there are many regional names in the DB which are not normalized correctly. However, having consulted with Jack the current changes are sufficient for him, so approving.

iankigen (approved):
Code/QA looks good
- Disparate values no longer exist for the same intended value in this field. ✅ 
- should be done via a data migration. ✅

PR #416, opened 2021-06-03
[AC-8759] Make CoreProfile into "socket layer" for Expert and Entrepreneur and MemberProfile models

iankigen:
[AC-8759](https://masschallenge.atlassian.net/browse/AC-8759)

**Changes introduced**
 - make CoreProfile into "socket layer" for Expert and Entrepreneur and MemberProfile models
 
 
**Testing**
- Confirm that:
  - an ObjectManager exists which is capable of returning all Profiles (Expert, Entrepreneur, or Member) which match a specific queryset
  - tests for CoreProfile demonstrate the use of the polymorphic functionality 
  
[sibling PR](https://github.com/masschallenge/accelerate/pull/2984)


iankigen:
moving expert profile test from accelerate to django-accelerator to address test coverage % drop

jonkiparsky:
We're going to need to convert the MemberProfiles as well as the Expert and Entrepreneurs - this should probably be added so that Jack can get a better sense of the new data picture.

jonkiparsky:
@shanbady your review is needed here as well

shanbady (approved):

jonkiparsky (approved):

PR #420, opened 2021-06-10
[AC-8836] Do not set last_login on user creation

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-8836

- verify newly created user has last_login of "None"


kenware (approved):
Looks good

iankigen (approved):
Code/QA looks good
- The last_login field displays the last date and time a user logged in ✅ 
- The last_login field is null when a user has never logged in ✅

PR #423, opened 2021-06-30
[AC-8974]: Add user note models

kenware:
[AC-8974](https://masschallenge.atlassian.net/browse/AC-8974)

How to test:
- Test is passing
- Definition of done matches [AC-8974](https://masschallenge.atlassian.net/browse/AC-8974)
- Check sibling PR https://github.com/masschallenge/accelerate/pull/3016


jonkiparsky:
One high-level issue here: the name of this model is "Note" but I suspect we're going to want to be more specific. In the future we're likely to want to offer notes about Startups, Partners, and other sorts of objects that may not even exist in the system now. 

I would ask that this class be called "UserNote" to allow room in the "Note" namespace for other specific note types, or to migrate this to a more generic "Note" class.

accelerator/tests/factories/note_factory.py
```diff
@@ -0,0 +1,18 @@
...
+from factory import Sequence, SubFactory
+
+from factory.django import DjangoModelFactory
+
+from simpleuser.tests.factories.user_factory import UserFactory
+
+Note = swapper.load_model('accelerator', 'UserNote')
+
+
+class NoteFactory(DjangoModelFactory):
```
jonkiparsky, line 12: rename as UserNoteFactory, consistent with model name

accelerator/models/user_note.py
```diff
@@ -0,0 +1,9 @@
+import swapper
+from accelerator_abstract.models import BaseUserNote
+
+
+class UserNote(BaseUserNote):
+    class Meta(BaseUserNote.Meta):
+        swappable = swapper.swappable_setting(
+            BaseUserNote.Meta.app_label, "Note"
```
iankigen, line 8: `"Note"` should be updated to `UserNote`

kenware: Thought I pushed

accelerator_abstract/models/base_user_note.py
```diff
@@ -0,0 +1,23 @@
...
+from django.db import models
+
+from accelerator_abstract.models.accelerator_model import AcceleratorModel
+
+
+class BaseUserNote(AcceleratorModel):
+    user = models.ForeignKey(settings.AUTH_USER_MODEL,
+                             related_name="user_notes",
+                             on_delete=models.CASCADE)
+    manager = models.ForeignKey(settings.AUTH_USER_MODEL,
```
iankigen, line 12: this should be renamed to `author`

iankigen (approved):
Code/QA looks good

accelerator/tests/factories/user_note_factory.py
```diff
@@ -0,0 +1,18 @@
+import swapper
+
+from factory import Sequence, SubFactory
```
jonkiparsky, line 3: multiple imports should be on separate lines
```
from factory import (
    Sequence, 
    SubFactory,
)
```

accelerator/tests/factories/user_note_factory.py
```diff
@@ -0,0 +1,18 @@
+import swapper
```
jonkiparsky, line 1: @iankigen is it possible that we can use what we learned from CAM Phase 1 and avoid Swappering this new model?

accelerator/tests/factories/user_note_factory.py
```diff
@@ -0,0 +1,18 @@
...
+
+from factory import Sequence, SubFactory
+
```
jonkiparsky, line 4: I would typically not have line breaks between all imports. It's nice to have a separation between "their" imports and "ours", but not between each line of imports.

jonkiparsky (approved):
This is basically fine - a few relatively trivial style notes which I would like to see addressed, but it seems like we're basically there. 

(but note change suggested on the sibling PR)

accelerator/tests/factories/user_note_factory.py
iankigen: yeah, it should be possible to avoid using swapper.

iankigen (commented):
Looks good

jonkiparsky:
verified unswapperized models work in django shell.

PR #431, opened 2021-08-25
[AC-9098] 0. Create PartnerLabel model

shakirandagire:
## [AC-9098](https://masschallenge.atlassian.net/browse/AC-9098)

**Changes introduced**
- Added PartnerLabel model

**How to test**
- Ensure that all the changes stated in [AC-9098](https://masschallenge.atlassian.net/browse/AC-9098) are met.


jonkiparsky (changes requested):
Please address linting on the migration, otherwise this looks good to me.

accelerator_abstract/models/base_partner_label.py
```diff
@@ -0,0 +1,24 @@
...
+from __future__ import unicode_literals
+
+import swapper
+from django.db import models
+
+from accelerator_abstract.models.label_model import LabelModel
+
+
+class BasePartnerLabel(LabelModel):
+    label = models.CharField(max_length=LabelModel.LABEL_LENGTH)
```
AmosWels, line 13: changes look good to me, the only observation is about the LABEL_LENGTH that is set to 255, for startups ad now partner label, would it be appropriate to separate the two lengths at this point or there won't be any foreseeable need since it's appropriate for labels

PR #438, opened 2021-09-10
[AC-9141] Update PartnerStartupInterest to reference Application rather than Startup

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-9141

Verify that model is updated as requested in the ticket


AmosWels (approved):
Changes meet the def of done,

The model has been changed from `PartnerStartupInterest` to `PartnerApplicationInterest`. The foreign key has been changed to Application, replacing the fk to Startup

shakirandagire (approved):
QA looks good me

PR #447, opened 2021-11-10
[AC-9123] Remove Finalist tag from all staff accounts

AmosWels:
Changes introduced in [AC-9123](https://masschallenge.atlassian.net/browse/AC-9123)

Remove Finalist tag from all staff accounts

**How to test:**

- Check out this branch
- Run Migrations on your localhost
- Notice that no staff user has a finalist role grant


shakirandagire (approved):
Code changes look good to me

accelerator/migrations/0075_auto_20211112_1309.py
```diff
@@ -0,0 +1,31 @@
...
+
+def remove_finalist_role_from_staff(apps, schema_editor):
+    ProgramRoleGrant = apps.get_model("accelerator", "ProgramRoleGrant")
+    staff_ids = User.objects.filter(
+        programrolegrant__program_role__user_role__name='Finalist',
+        is_staff=True).values_list('id', flat=True)
+
+    staff_finalist_role = ProgramRoleGrant.objects.filter(
+                program_role__user_role__name='Finalist',
+                person_id__in=staff_ids).values_list(
```
iankigen, line 15: instead of getting the id values list then deleting each program role grant in a loop, it is much more efficient to do this instead:
```
ProgramRoleGrant.objects.filter(
                program_role__user_role__name='Finalist',
                person_id__in=staff_ids).delete()

**Review by iankigen — CHANGES_REQUESTED** [review]

File: accelerator/migrations/0075_auto_20211112_1309.py [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+
+from django.db import migrations
+from simpleuser.models import User
+
+
+def remove_finalist_role_from_staff(apps, schema_editor):
+    ProgramRoleGrant = apps.get_model("accelerator", "ProgramRoleGrant")
+    staff_ids = User.objects.filter(
+        programrolegrant__program_role__user_role__name='Finalist',
+        is_staff=True).values_list('id', flat=True)
```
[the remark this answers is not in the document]
AmosWels: thanks, let me update this query

accelerator/migrations/0075_auto_20211112_1309.py
```diff
@@ -0,0 +1,32 @@
...
+
+def remove_finalist_role_from_staff(apps, schema_editor):
+    ProgramRoleGrant = apps.get_model("accelerator", "ProgramRoleGrant")
+    staff_ids = []
+    finalist_users = User.objects.filter(
+              programrolegrant__program_role__user_role__name='Finalist')
+
+    for user in finalist_users:
+        if is_employee(user):
+            staff_ids.append(user.id)
```
iankigen, line 16: instead of doing this with a loop which is always very expensive
we can have
```
staff_ids = set(Clearance.objects.all().values_list('user__pk', flat=True)).union(set(User.objects.filter(is_superuser=True).values_list('pk', flat=True)))

**Comment by iankigen** [comment]
we have a failing build

File: accelerator/migrations/0075_auto_20211112_1309.py (hunk 3 above)
Comment (AmosWels) [reply accelerator/migrations/0075_auto_20211112_1309.py:16]: this gets rid of the `is_employee` check that was earlier suggested

Comment (iankigen) [reply accelerator/migrations/0075_auto_20211112_1309.py]: def is_employee(user) was meant to be used as a reference on how to get a staff user

**Comment by AmosWels** [comment]
> we have a failing build

>Noticed this failing build and tried resolving it with the git config though it persists. Any pointers are welcome

**Comment by jonkiparsky** [comment]
> > we have a failing build
> 
> > Noticed this failing build and tried resolving it with the git config though it persists. Any pointers are welcome

Since this is running under travis, local git config changes won't fix it. 

I'm trying to understand what the issue is, not sure yet. It first appears at https://github.com/masschallenge/django-accelerator/pull/447/commits/d9fe970a1a66b60017832afbf07931dc89037c70 which seems clean.

File: accelerator/migrations/0075_auto_20211112_1309.py [hunk 4]
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 2.2.10 on 2021-11-12 18:09
```
jonkiparsky, line 1: @AmosWels Can you please give this migration a descriptive name?

accelerator/migrations/0075_auto_20211112_1309.py
```diff
@@ -0,0 +1,27 @@
...
+
+from django.db import migrations
+from simpleuser.models import User
+
+
+def remove_finalist_role_from_staff(apps, schema_editor):
+    ProgramRoleGrant = apps.get_model("accelerator", "ProgramRoleGrant")
+    Clearance = apps.get_model("accelerator", "Clearance")
+    staff_ids = set(Clearance.objects.all().values_list(
```
kenware, line 10: This query takes all user that has any clearance as staff. @iankigen advice if I am wrong.

jonkiparsky: @kenware That is correct. Users who have Clearances are presumed to be staff.

kenware: @AmosWels you need to filter users with staff clearances and do a union with superusers.

PR #453, opened 2021-11-23
[AC-9270] Update JudgeProgressReport to fetch judge info from CoreProfile model

iankigen:
## [AC-9270](https://masschallenge.atlassian.net/browse/AC-9270)

**Changes introduced**
- move functions from expert profile to core profile

[Sibling PR](https://github.com/masschallenge/accelerate/pull/3138)


jonkiparsky (approved):
Pretty straightforward, looks good.

accelerator/models/expert_profile.py
```diff
@@ -90,132 +73,18 @@ def has_invited_judging_rounds(self):
...
-        roles = role_dict.get(state, [UserRole.DESIRED_JUDGE, UserRole.JUDGE])
-        if program:
-            return self.user.programrolegrant_set.filter(
-                program_role__program__exact=program,
-                program_role__user_role__name__in=roles).exists()
-        else:
-            return self.user.programrolegrant_set.filter(
-                program_role__user_role__name__in=roles).exists()
-
     def get_active_alerts(self, page=None):
```
AmosWels, line 88: looking at some class methods that are moved to the core profile like `get_active_alerts`, and if these won't be required to be moved later

iankigen: the methods were not moved since they are not directly related to functionality delivered by this ticket

AmosWels (approved):
code changes look good to me

PR #466, opened 2022-01-31
[AC-9364] Add a new field to the CoreProfile and create a CommunityParticipation model

iankigen:
## [AC-9364](https://masschallenge.atlassian.net/browse/AC-9364)

**Changes introduced**
- Add CommunityParticipation model

**Testing**
- Confirm that the requirements on [AC-9364](https://masschallenge.atlassian.net/browse/AC-9364) are satisfied


accelerator/tests/factories/core_profile_factory.py
```diff
@@ -74,3 +74,11 @@ def ethno_racial_identification(self, create, extracted, **kwargs):
         if extracted:
             for tag in extracted:
                 self.ethno_racial_identification.add(tag)
+
+    @post_generation
+    def community_participation(self, create, extracted, **kwargs):
```
kenware, line 79: At what point do you expect this hook to run?

kenware: I imagine adding participation to a user will happen at the Django form

iankigen: this will be used to generate community_participation test data. 
https://factoryboy.readthedocs.io/en/stable/recipes.html?highlight=post_generation#simple-many-to-many-relationship

kenware:
Could you fix the failing test

kenware:
> could we be missing an abstract base class for the CommunityParticipation model?

That is a good point @AmosWels . @iankigen any reason we are not having base classes for community participation?

iankigen:
> could we be missing an abstract base class for the CommunityParticipation model?

it is not necessary since we are not using swappable on this model

kenware (approved):
This looks good

AmosWels (approved):
changes look good and migrations run succesfully

shakirandagire (approved):
changes Iook good to me

PR #467, opened 2022-01-31
[AC-8938]: update core profile models

kenware:
[AC-8938](https://masschallenge.atlassian.net/browse/AC-8938)

How to test:
- Ensure all the field and properties are added on the core profile


iankigen:
we migration conflicts

accelerator_abstract/models/base_core_profile.py
```diff
@@ -305,6 +341,43 @@ class BaseCoreProfile(AcceleratorModel):
         verbose_name="Expert Group",
         max_length=10,
         blank=True)
+    preferred_name = models.CharField(
+        verbose_name="Prefered Name",
```
iankigen, line 345: according to mockup this should be: `Nickname / Preferred Name`

accelerator_abstract/models/base_core_profile.py
```diff
@@ -305,6 +341,43 @@ class BaseCoreProfile(AcceleratorModel):
...
+        max_length=32,
+        blank=True)
```
iankigen, line 347: null should also be set to True. Default is False.

accelerator_abstract/models/base_core_profile.py
```diff
@@ -305,6 +341,43 @@ class BaseCoreProfile(AcceleratorModel):
...
+    pronouns = models.CharField(
+        verbose_name="Prefered Name",
```
iankigen, line 349: this seems wrong

accelerator_abstract/models/base_core_profile.py
```diff
@@ -305,6 +341,43 @@ class BaseCoreProfile(AcceleratorModel):
...
+        max_length=32,
+        choices=PRONOUN_CHOICES,
+        blank=True)
```
iankigen, line 352: null should also be set to True. Default is False. (do the same for all the fields since they are all optional)

accelerator_abstract/models/base_core_profile.py
```diff
@@ -305,6 +341,43 @@ class BaseCoreProfile(AcceleratorModel):
...
+    education_level = models.CharField(
+        verbose_name="Educational Level",
```
iankigen, line 354: should be `Education Level`

accelerator_abstract/models/base_core_profile.py
```diff
@@ -87,6 +87,42 @@
                       "as how you think they might want to be involved "
                       "(Judge, Mentor, etc.) Also, please encourage these "
                       "individuals to fill out their own Expert Profile.")
+PRONOUN_CHOICES = (
+    ('she, her, hers', 'She, Her, Hers'),
```
iankigen, line 91: It would be nice to have a suitably-named constant for each value to make it easy to import the choice value

iankigen (commented):
I also noticed the core profile factory was not updated with the new fields

kenware:
> 

I have updated the related model `user_location`. For other fields, since all can be nullable. It is not necessary. It is a model factory so you can always pass in those field with there value when using the factory

kenware:
>

accelerator_abstract/models/base_core_profile.py
```diff
@@ -305,6 +354,56 @@ class BaseCoreProfile(AcceleratorModel):
...
+        verbose_name="Enterpreneur Interest",
+        null=True,
+        default=False)
+    worldwide_participation_interest = models.BooleanField(
+        verbose_name="World Wide Participation Interest",
+        null=True,
+        default=False)
+    shared_demographic_data = models.BooleanField(
+        verbose_name="Permission To Shared Demographic Data",
+        null=True,
```
iankigen, line 399: not necessary since you are providing a default(applies to all boolean fields)

iankigen (approved):
Code/QA looks good
- specified fields have been added to the CoreProfile model

shakirandagire (approved):
QA looks good to me

PR #468, opened 2022-02-01
[AC-8287] Eliminate RemovedInDjango31Warning: Startup QuerySet won't use Meta.ordering in Django 3.1.

AmosWels:
[AC-8287](https://masschallenge.atlassian.net/browse/AC-8287)

How to test:
- Ensure that meta ordering is removed from the startup model


accelerator/migrations/0079_auto_20220201_0117.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.2.10 on 2022-02-01 06:17
```
iankigen, line 1: can you rename the migration file to something descriptive

iankigen (commented):
The string `QuerySet won't use Meta.ordering in Django 3.1.` does not appear when running tests ✅

shakirandagire (approved):
Code changes look good to me.

iankigen (approved):

PR #476, opened 2022-02-24
[AC-9340] Create Personal Profile progress bar

iankigen:
## [AC-9340](https://masschallenge.atlassian.net/browse/AC-9340)

**Changes introduced**
- add percent_complete function to get profile completion progress in %

**Testing**
- Visit any of the unified profile views e.g https://test-cam.masschallenge.org/profile/ 
- Confirm that the definition of done according to [AC-9340](https://masschallenge.atlassian.net/browse/AC-9340) is satisfied

[Sibling PR](https://github.com/masschallenge/accelerate/pull/3196)

### DIFF
--- a/accelerator/models/core_profile.py
+++ b/accelerator/models/core_profile.py
@@ -42,6 +42,19 @@ OFFICE_HOURS_HOLDER = Q(
 ACTIVE_PROGRAM = Q(
     program_role__program__program_status=ACTIVE_PROGRAM_STATUS)
 
+PROFILE_USER_FIELDS = ['first_name', 'last_name']
+PROFILE_FIELDS = [
+    'shared_demographic_data', 'education_level', 'privacy_email',
+    'privacy_web', 'privacy_profile', 'primary_industry',
+    'functional_expertise', 'additional_industries', 'privacy_phone',
+    'linked_in_url', 'geographic_experience', 'program_families',
+    'authorization_to_share_ethno_racial_identity']
+PROFILE_LOCATION_FIELDS = [
+    'street_address', 'city', 'state', 'country']
+COMMUNITY_PARTICIPATION_TYPES = [
+    PARTICIPATION_ROLE, ADVISING_STARTUP_BUSINESS_AREA,
+    ADVISING_STARTUP_SUPPORT_BELIEVE]
+
 
 class CoreProfile(BaseCoreProfile, PolymorphicModel):
     class Meta(BaseCoreProfile.Meta):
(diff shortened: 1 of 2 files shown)


accelerator/models/core_profile.py
```diff
@@ -386,6 +392,37 @@ def is_judge(self, program=None, state=None):
             return self.user.programrolegrant_set.filter(
                 program_role__user_role__name__in=roles).exists()
 
+    @property
+    def percent_complete(self):
+        completed_count = 0
+        profile_user_fields = ['first_name', 'last_name']
+        profile_fields = [
+            'shared_demographic_data', 'education_level', 'privacy_email',
+            'privacy_web', 'privacy_profile', 'primary_industry',
```
kenware, line 401: There seem to be a missing fields here. like bio, gender identity etc

iankigen: using the list from https://masschallenge.atlassian.net/browse/AC-9340

accelerator/tests/test_core_profile.py
```diff
@@ -368,6 +376,20 @@ def was_mentor_in_last_12_months(self):
         _user_with_role(profile.user, BaseUserRole.MENTOR, program=program)
         self.assertTrue(profile.was_mentor_in_last_12_months())
 
+    def test_completion_percentage_is_correct_for_completed_profile(self):
```
AmosWels, line 379: it would be good to add a test to the case for incomplete profile

accelerator/models/core_profile.py
```diff
@@ -386,6 +392,37 @@ def is_judge(self, program=None, state=None):
...
+        community_participation_types = [
+            PARTICIPATION_ROLE, ADVISING_STARTUP_BUSINESS_AREA,
+            ADVISING_STARTUP_SUPPORT_BELIEVE]
+        profile_data_dict = model_to_dict(self, profile_fields)
+        user_data_dict = model_to_dict(self.user, profile_user_fields)
+        completed_participation_by_type = self.community_participation.filter(
+            type__in=community_participation_types).values_list(
+            'type', flat=True).distinct().count()
+        if self.user_location:
+            location_data_dict = model_to_dict(
```
AmosWels, line 416: Non-blocker: 
wouldn't the `location_data_dict` be counted as part of the total when when `user_location` is `True` ?

iankigen: `profile_location_fields` is used when counting total. len(specified fields) = total

AmosWels (approved):
changes looking good to me

accelerator/models/core_profile.py
```diff
@@ -386,6 +392,37 @@ def is_judge(self, program=None, state=None):
             return self.user.programrolegrant_set.filter(
                 program_role__user_role__name__in=roles).exists()
 
+    @property
```
jonkiparsky, line 395: Not a blocker, but I'm not keen on the use of the `@property` decorator here. Functions tend to acquire parameters, and making this a property makes it harder to add params later. 

But since we're under time pressure, let's not worry about this.

accelerator/models/core_profile.py
```diff
@@ -386,6 +392,37 @@ def is_judge(self, program=None, state=None):
...
+    def percent_complete(self):
+        completed_count = 0
+        profile_user_fields = ['first_name', 'last_name']
+        profile_fields = [
```
jonkiparsky, line 399: I would prefer to have this list (and the others like it) declared outside of the function. Among other things, this allows reuse of the list, it focuses the reader's attention on the logic rather than on this data, and incidentally it shortens the function and brings it closer to our line limits :)

accelerator/models/core_profile.py
```diff
@@ -386,6 +392,37 @@ def is_judge(self, program=None, state=None):
...
+        if self.user_location:
+            location_data_dict = model_to_dict(
+                self.user_location, profile_location_fields)
+            completed_count += _get_completed_fields_count(location_data_dict)
+        completed_count += _get_completed_fields_count(profile_data_dict)
+        completed_count += _get_completed_fields_count(user_data_dict)
+        completed_count += completed_participation_by_type
+        total = len(community_participation_types + profile_fields
+                    + profile_user_fields + profile_location_fields)
+        return completed_count / total * 100
```
jonkiparsky, line 424: The ticket specifies 
>returns a number between zero and 1 (inclusive) indicating the percentage of “interesting fields” that the user has filled in

This looks like it returns a number in the range 0..100.

kenware (approved):
Looks good

PR #481, opened 2022-03-03
[AC-9381]: Create bussiness proposition model

kenware:
[AC-9381](https://masschallenge.atlassian.net/browse/AC-9381l)


accelerator_abstract/tests/test_bussiness_proposition.py
```diff
@@ -0,0 +1,27 @@
+from django.test import TestCase
+
+from accelerator.tests.factories import (
+    BusinessPropositionFactory
+)
+from accelerator.models import BusinessProposition
+from accelerator_abstract.models import EXCLUDED_FIELDS
```
iankigen, line 7: import order

accelerator_abstract/models/base_business_proposition.py
```diff
@@ -0,0 +1,86 @@
+from __future__ import unicode_literals
```
iankigen, line 1: we don't need this(same applies for other files with unicode_literals)

accelerator/migrations/0087_create_business_proposition_20220303_0338.py
```diff
@@ -0,0 +1,107 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accelerator', '0086_alter_founder_field'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='coreprofile',
+            name='preferred_name',
```
iankigen, line 18: why do we have coreprofile field change?

iankigen (commented):
For the most part, QA/Code changes look good
- A new BusinessProposition model is added to the accelerate DB. This model includes the fields described in the mockup (listed below), and has a method `complete` which returns True iff all associated fields have at least 20 characters of text.  ✅ 
- The `BusinessProposition` has a foreign-key relationship to the Startup model ✅ 
- Data fields of the BusinessProposition model ✅ 

From the mockup, `BusinessProposition` model seems to have help text for each field. was that left out intentionally ❓ 
We have

We have lint warnings

accelerator_abstract/models/base_business_proposition.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __str__(self):
+        return self.startup.organization.name
+
+    def complete(self):
+        fields = self._meta.get_fields(include_parents=False)
+        for field in fields:
+            is_text_field = field.name not in EXCLUDED_FIELDS
+            value = getattr(self, field.name)
+            if is_text_field and (not value or len(value) < 20):
```
AmosWels, line 82: I think this check should be `<= 20`

kenware: The ticket says it should return true if it is 20 or above(at least 20 characters). `<= 20` will mean it should return true if it is above 20 characters.

`and has a method complete which returns True iff all associated fields have at least 20 characters of text. `

AmosWels: right, i had considered the inverse

AmosWels (approved):
migrations run successfully and changes look good to me

PR #490, opened 2022-03-23
[AC-9358]: Company Profile Progress indicator

kenware:
[AC-9358](https://masschallenge.atlassian.net/browse/AC-9358)

How to test:
- On your local machine, run `django-shell`
- Fetch any startup instance and call `startup.progress_complete()`
- You can save more data for for that instance based on the specified fields on [AC-9358](https://masschallenge.atlassian.net/browse/AC-9358) for either startup or business proposition to see how the progress increases and milestones change

NB: A quicker approach to testing is the follow my test cases and run in django-shell using test factories for all possible scenarios


accelerator_abstract/models/base_startup.py
```diff
@@ -272,3 +292,57 @@ def is_finalist(self, program=None):
         ).exists()
 
     is_finalist.boolean = True
+
+    def progress_complete(self):
```
AmosWels, line 296: you can name this method as `profile_status`

kenware: Other methods on the application that has to do with progress are named this way, is there special reason we should change it to `profile_status`

accelerator_abstract/utils.py
```diff
@@ -32,4 +32,20 @@ def _error_msg(e):
     return msg
 
 
+def _calc_progress(total, progress_num, **kwargs):
+    return {'progress': round(progress_num/total, 2),
```
AmosWels, line 36: why not round off to 1

AmosWels (commented):
QA looks good,  a dict is returned after filling out the fields
```
{'progress': 1.0,
 'milestone': 'complete-profile',
 'profile-complete': True,
 'bus-prop-complete': True}

File: accelerator_abstract/models/base_startup.py (hunk 1 above)
Comment (AmosWels) [reply accelerator_abstract/models/base_startup.py:296]: it was specified in the def of done. If this is done for consistency, then we may consider updating the def of done

File: accelerator_abstract/utils.py (hunk 2 above)
Comment (AmosWels) [reply accelerator_abstract/utils.py:36]: right, and that would be 0.7 if rounded off to 1 dp

Comment (kenware) [reply accelerator_abstract/utils.py:36]: Rounding off to 1dp will always give the profile progress in a multiple of 10. like 10%, 20% 30% etc. Does not represent an accurate increment from me. That means you can fill some field without detecting any change till it jumps across the next round of multiple of 10 %.

Comment (kenware) [reply accelerator_abstract/utils.py:36]: I would rather have profile increase as 61 => 62 => 63=>64=>65 =>66=>67 etc.
instead of  60 => 70 => 80 => 90 => 100(This is the result rounding to 1 dp will give).

File: accelerator_abstract/models/base_startup.py [hunk 3]
```diff
@@ -272,3 +292,56 @@ def is_finalist(self, program=None):
...
+
+        instance = self.business_propositions.order_by('created_at').last()
+        bus_prop_fields = _get_model_fields(self.business_propositions.model,
+                                            EXCLUDED_FIELDS)
+        total_fields = len(STARTUP_FIELDS) + len(bus_prop_fields)
+
+        prof_progress_num, prof_milestone, profile = self._field_to_data(
+            self,
+            STARTUP_FIELDS)
+        bus_p_progress_num, bus_p_milestone, bus_p = self._field_to_data(
```
shakirandagire, line 307: Some names are a bit confusing like bus_p

AmosWels (approved):
changes look good

shakirandagire (approved):
Changes look good to me

PR #491, opened 2022-03-23
[AC-9467]: Make Startup fields nullable

AmosWels:
[AC-9467](https://masschallenge.atlassian.net/browse/AC-9467)

How to test:
- On your local machine, run migrations
- Check that the def of done according to [AC-9467](https://masschallenge.atlassian.net/browse/AC-9467) is satisfied


accelerator/migrations/0093_alter_startup_organization_fields.py
```diff
@@ -0,0 +1,632 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accelerator', '0092_update_business_proposition_20220318_1245'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='allocator',
```
iankigen, line 17: we have multiple model options alterations that are not related to the Startup model. Why is that the case?

AmosWels: changes were only made to the startup and organization model. I'm not certain how this change got Introduced

iankigen: I think we should remove the model options alterations unless it was intentional

accelerator/migrations/0094_alter_startup_organization_fields.py
```diff
@@ -0,0 +1,305 @@
+# Generated by Django 2.2.27 on 2022-03-25 08:52
+
+from django.conf import settings
+import django.core.validators
+from django.db import migrations, models
+import django.db.models.deletion
+import embed_video.fields
+import sorl.thumbnail.fields
+
```
iankigen, line 9: import order
```
import django.core.validators
import django.db.models.deletion
import embed_video.fields
import sorl.thumbnail.fields
from django.conf import settings
from django.db import migrations, models

iankigen (approved):
QA/Code looks good
- All fields of the Startup and Organization models are nullable (blank=True, null=True) except for organization name

shakirandagire (approved):
Code changes look good to me

PR #492, opened 2022-03-29
[AC-9476] Make judging dashboard available to all users in side nav

iankigen:
## [AC-9476](https://masschallenge.atlassian.net/browse/AC-9476)

**Changes Introduced**
- Update judging dashboard nav items access restrictions to allow all users regardless of their user_type or user_role

**Testing**
- Confirm that the following conditions are satisfied:
  - All users can see the “Judging” tab in their side nav, regardless of user type or program role grants
  - Users clicking on the “Judging” can access “Judging Commitments” and “Judging Panels” without error

Test cases
- Expert|Entrepreneur|Unified user with+without a user_role, with+without commitments (save commitment), with+without Judge panel assignment

URL: https://test-ikigen.masschallenge.org/


phembarl:
While on the `Judging Panels` tab, the side nav is not visible. Is that the expected behaviour?

phembarl:
The side nav interface is different when on the `Judging` section compared to the other sections (Home, Mentoring, Companies). Kindly confirm if this is the expected behaviour as well

iankigen:
>While on the Judging Panels tab, the side nav is not visible. Is that the expected behaviour?

as it stands, the side nav is disabled in the Judging Panels view (current behavior)

iankigen:
>The side nav interface is different when on the Judging section compared to the other sections (Home, Mentoring, Companies). Kindly confirm if this is the expected behaviour as well

The judging Overview interface is using semantic UI style while the other interfaces are using bootstrap theme

phembarl (approved):

kenware (commented):
Is it possible to add test cases for this change?

iankigen:
I'm not sure writing unit test for data migration is a good idea

kenware:
> I'm not sure writing unit test for data migration is a good idea

I was referring to the test on the view about visibility but this is non-blocking

kenware:
> I'm not sure writing unit test for data migration is a good idea

I was referring to the test on the view about visibility but this is non-blocking

PR #520, opened 2022-06-20
[AC-9622]: [Program Interest]: Create New Industry field on Program Admin

AmosWels:
[AC-9622](https://masschallenge.atlassian.net/browse/AC-9622)

How to test
- Check out branch `AC-9622`
- Run `make migrate`
- Ensure that the def of done according to [AC-9622](https://masschallenge.atlassian.net/browse/AC-9622) are met

[Sibling PR](https://github.com/masschallenge/accelerate/pull/3319)


accelerator/migrations/0107_add_associated_industry_field.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.28 on 2022-06-20 15:16
+
+import django.core.validators
```
iankigen, line 3: Imported but unused

accelerator/migrations/0107_add_associated_industry_field.py
```diff
@@ -0,0 +1,19 @@
...
+from django.db import migrations, models
```
iankigen, line 4: can we have this as a multiline import?

iankigen (commented):
We have migration conflicts 
Otherwise, looks good

iankigen (approved):

accelerator_abstract/models/base_program.py
```diff
@@ -160,6 +160,10 @@ class BaseProgram(AcceleratorModel):
         swapper.get_model_name(
             AcceleratorModel.Meta.app_label, 'IndustryCluster'),
         blank=True, related_name='programs')
+    associated_industry = models.CharField(
+        max_length=20,
```
kenware, line 164: How can we determine that max length will always be less or equal to 20? I would suggest 50 might be more save

AmosWels: 50 might be a bit too long for the display available. 

During estimation, @PrinceAdeW mentioned that the 20 was specified in the ticket with consideration of the display as well.

kenware (approved):
Looks good

PR #533, opened 2022-07-22
AC-9537 [Program Interest] Changes on Program categories Section

iankigen:
## [AC-9537](https://masschallenge.atlassian.net/browse/AC-9537)

**Changes introduced**
- add geographical ecosystem model

**Testing**
- On accelerate, `make migrate`
- Visit http://localhost:8181/profile/programs/ and http://localhost:8181/admin/accelerator/ecosystem/
- Confirm that the def of done according to [AC-9537](https://masschallenge.atlassian.net/browse/AC-9537) is satisfied

[Sibling PR](https://github.com/masschallenge/accelerate/pull/3364)


accelerator/tests/factories/ecosystem_factory.py
```diff
@@ -0,0 +1,13 @@
...
+from accelerator.models import Ecosystem
+
+
+class EcosystemFactory(DjangoModelFactory):
+    class Meta:
+        model = Ecosystem
+        django_get_or_create = ['name']
+
+    name = Sequence(lambda n: "Ecosystem {}".format(n))
+    display_priority = Sequence(lambda n: n)
```
AmosWels, line 13: could you include the `is_visible` field to the factory

AmosWels (approved):
changes look good to me and migrations run successfully as per the def of done

AmosWels (commented):
We have a failing build

accelerator/tests/factories/ecosystem_factory.py
iankigen: it is included by default and since it already has a default of True - there's no need to repeat the same

shakirandagire (approved):
Code changes look good to me

